PR #49, opened 2017-06-18
Remove Fortran95Experimental

GuiltyDolphin:
We have active Fortran95 support being developed
on https://github.com/camfort/fortran-src/tree/hc-parser-95.

Having the experimental Fortran95 support on master just increases build times, we should just take it out until support is finalized.


dorchard:
Ok I think this is fine. We don't depend on it in any of our paper results.

PR #52, opened 2017-07-06
Add Fortran 95 support to readme

GuiltyDolphin:
@harry-clarke Just specified that we have (after this is complete) support for Fortran 95 in readme.


dorchard:
Am I right in thinking that this pull request is out of date and is subsumed by the commits @harry-clarke did into master? If so I can delete it.

PR #57, opened 2017-12-12
rGetDirContents should track directories it has seen

gridaphobe:
This avoids infinite loops due to symlinks pointing back up the directory tree.

As an example, try adding a symlink from `fortran-src/src` to `fortran-src`, e.g.

```
ln -s /path/to/fortran-src /path/to/fortran-src/src/loop
```


dorchard:
Great- thanks!

PR #58, opened 2018-01-24
extend Fortran77 lexer/parser with support for legacy extensions

gridaphobe:
- TYPE statement (alias for PRINT)
- STRUCTURE/UNION/MAP statements
- hollerith constants
- BOZ literals
- inline comments (we just throw them away for now)
- CASE statement
- specific widths for type declarations (eg, INTEGER*4)

### DIFF
--- a/src/Language/Fortran/Lexer/FixedForm.x
+++ b/src/Language/Fortran/Lexer/FixedForm.x
@@ -1,5 +1,6 @@
 -- -*- Mode: Haskell -*-
 {
+{-# LANGUAGE BangPatterns #-}
 {-# LANGUAGE FlexibleInstances #-}
 {-# LANGUAGE DeriveGeneric #-}
 {-# LANGUAGE DeriveDataTypeable #-}
@@ -7,12 +8,15 @@
 {-# LANGUAGE FlexibleContexts #-}
 {-# LANGUAGE UndecidableInstances #-}
 
-module Language.Fortran.Lexer.FixedForm where
+module Language.Fortran.Lexer.FixedForm
+  ( lexer, initParseState, collectFixedTokens, collectFixedTokensSafe
+  , Token(..), LexAction, AlexInput(..), lexemeMatch, lexN
+  ) where
 
 import Data.Word (Word8)
-import Data.Char (toLower, ord)
+import Data.Char (toLower, ord, isDigit)
 import Data.List (isPrefixOf, any)
-import Data.Maybe (fromJust, isNothing)
+import Data.Maybe (fromJust, isNothing, isJust)
 import Data.Data
 import qualified Data.Bits
 import qualified Data.ByteString.Char8 as B
@@ -29,6 +33,14 @@ import Language.Fortran.Util.Position
 }
 
 $digit = [0-9]
+$octalDigit = 0-7
+$hexDigit = [a-f $digit]
+$bit = 0-1
+
+@binary = b\'$bit+\' | \'$bit+\'b
+@octal = o\'$octalDigit+\' | \'$octalDigit+\'o
+@hex = x\'$hexDigit+\' | \'$hexDigit+\'x | z\'$hexDigit+\' | \'$hexDigit+\'z
+
 $letter = [a-z]
 $alphanumeric = [$letter $digit]
 $alphanumericExtended = [$letter $digit \_]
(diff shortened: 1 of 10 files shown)


src/Language/Fortran/Lexer/FixedForm.x
```diff
@@ -162,20 +216,24 @@ tokens :-
...
+  <st,iif> ".xor." / { legacy77P }            { addSpan TOpXOr  }
   <st,iif> ".eqv." / { fortran77P }           { addSpan TOpEquivalent  }
   <st,iif> ".neqv." / { fortran77P }          { addSpan TOpNotEquivalent  }
 
   -- Relational operators
   <st,iif> "<" / { extended77P }              { addSpan TOpLT  }
   <st,iif> "<=" / { extended77P }             { addSpan TOpLE  }
   <st,iif> "==" / { extended77P }             { addSpan TOpEQ  }
-  <st,iif> "!=" / { extended77P }             { addSpan TOpNE  }
+  <st,iif> "/=" / { extended77P }             { addSpan TOpNE  }
```
gridaphobe: As far as I can tell, `!=` is not a valid operator in f77. `gfortran` treats the `!` as starting an inline comment, and instead accepts `/=`.

src/Language/Fortran/Parser/Fortran77.y
```diff
@@ -136,6 +169,24 @@ import Debug.Trace
 
 %%
 
+maybe(p)
```
gridaphobe: turns out recent versions of happy *do* support parameterized productions! 

https://www.haskell.org/happy/doc/html/sec-grammar.html

gridaphobe:
Sorry in advance, I know this is going to be a hard patch to review..

dorchard:
Fantastic- thanks so much. We'll try to sit down soon to review this! Do you think it will require any changes to CamFort as well?

gridaphobe:
Ah, yes, probably a couple. Off the top of my head:

1. I added a new [`byte`](https://docs.oracle.com/cd/E19957-01/805-4939/6j4m0vn7n/index.html) type, which should be interpreted as a 1-byte integer.
2. The [`structure`](https://docs.oracle.com/cd/E19957-01/805-4939/6j4m0vnbe/index.html) and [`union/map`](https://docs.oracle.com/cd/E19957-01/805-4939/6j4m0vnbl/index.html) statements will probably need camfort changes to understand the memory layout.

mrd:
Is it possible that we might combine the Fortran77Extended and Fortran77Legacy modes as a general 'extended and/or legacy' mode?

gridaphobe:
I would prefer that, actually. The main obstacle that comes to mind is comments. 

Fortran77Legacy omits comments because the addition of inline comments makes it very difficult to preserve them in the core AST. This means that the current refactoring support, based on pretty-printing the transformed AST, would destroy all inline comments, which is no good. As an alternative, we could rewrite only the portion of the AST that has changed, which would minimize the loss.

mrd:
If that's the obstacle then we could handle that by issuing an error when coming across inline comments, allowing it to be downgraded to a warning with a command line option. That at least get the patch integrated into the repository.

And then we start talking about how we would like to actually handle inline comments :)

mrd:
What's holding me up currently is a strange type error when compiling fortran-src as a dependency of camfort. This does not appear when compiling fortran-src alone.

Are you able to compile camfort? Getting a few dozen errors of the form:

```
Fortran77.hs:5524:19: error:
        • Couldn't match type ‘t131’ with ‘[Block A0]’
          ‘t131’ is a rigid type variable bound by
            the type signature for:
              happyReduce_418 :: forall t126 t127 t128 t129 t130 t131.
                                 Happy_GHC_Exts.Int#
                                 -> Token
                                 -> Happy_GHC_Exts.Int#
                                 -> Happy_IntList
                                 -> HappyStk (HappyAbsSyn t126 t127 t128 t129 t130 t131)
                                 -> LexAction (HappyAbsSyn t126 t127 t128 t129 t130 t131)
```

gridaphobe:
Hrm, I get the same error.. Very strange that compiling fortran-src as a dependency of camfort would fail.

gridaphobe:
It appears to be an issue of the version of happy (1.19.7) included in the stack 9.5 lts. The error doesn't happen in fortran-src because it's using an older lts with an older happy. If you move camfort to the newer 9.21 lts, with a newer happy, the error goes away. (You might need to do a `stack clean` to clear out the generated .hs file)

mrd:
That works thanks. I will go ahead and implement the proposed compromise and think about inline comments some more.

mrd:
Before I proceed to integrate the Legacy and Extended flavours -- do you have a set of sample Legacy77 Fortran files to use as testcases?

gridaphobe:
@mrd, sorry for the slow reply. I can't provide any of the actual files I tested against (they're proprietary code), but I'll try to put together a test file of my own this afternoon that covers the issues I remember running into.

Also more than happy to test new builds internally and send back patches (with test coverage this time!) for anything that breaks.

PR #64, opened 2018-03-05
add tests for Fortran77Legacy

gridaphobe:


gridaphobe:
Sorry to be so slow with this! I combed through my commits and the files I tested with, and extracted samples of the tricky statements. Seems to mostly be dealing with comments and non-standard type specifications.

mrd:
No problem, thanks!

PR #65, opened 2018-07-13
Fix pedantic warnings

traveltissues:
Adds missing signatures and replaces unused bound variables with anonymous ones. Also resolves some import warnings. No fixes for test specs included.


mrd:
Thanks

PR #66, opened 2018-07-19
Don't expand and reparse already expanded files

traveltissues:
Currently if `A` includes `{B,C}` and `B` includes `C` then `C` will be parsed twice inflating runtime costs. Suggestion to not reparse `C` but instead just return the statement.


mrd:
thanks

PR #67, opened 2018-07-22
Fix shadowing

traveltissues:
Resolves name shadowing throughout. This looks less elegant in some places but improves legibility in my opinion by making binds explicit.


mrd:
thanks

Issue #43, opened 2017-03-15
dataFlowSolver exception


dorchard:
The code in `computational-physics-1/fortran77/code214.f` raises a dataFlowSolver exception that I can't debug, see:  ```fortran-src code214.f -S```

dorchard:
The error here has gotten much better:
```
$ fortran-src computational-physics-1/fortran77/code214.f -S
fortran-src: An unexpected LHS to an expression assign: ExpFunctionCall () (42:8)-(42:14) (ExpValue () (42:8)-(42:12) (ValVariable "ranf")) Nothing
CallStack (from HasCallStack):
  error, called at src/Language/Fortran/Analysis.hs:239:16 in main:Language.Fortran.Analysis
```
It seems there is some syntactic construct which we are not expecting where line 42 is:
`        RANF() = ISEED/FLOAT(IC)`
(this is inside the definition of the function `RANF`). This is possibly non-standard Fortran so it is not our fault- but we should check the Fortran 77 standard to see if this is allowed or not.

mrd:
I think that it might be a typo.. Check out code213.f

traveltissues:
does #72 close this?

mrd:
I'm 100% certain this was a typo in code214.f, nothing to do with us.

      FUNCTION RANF()
      DATA IA/16807/,IC/2147483647/,IQ/127773/,IR/2836/
      COMMON /CSEED/ ISEED
        IH = ISEED/IQ
        IL = MOD(ISEED,IQ)
        IT = IA*IL-IR*IH
        IF(IT.GT.0) THEN
          ISEED = IT
        ELSE
          ISEED = IC+IT
        END IF
        RANF() = ISEED/FLOAT(IC)
      RETURN
      END

whereas the same exact function (but for one change) in code213.f:


      FUNCTION RANF()
      DATA IA/16807/,IC/2147483647/,IQ/127773/,IR/2836/
      COMMON /CSEED/ ISEED
        IH = ISEED/IQ
        IL = MOD(ISEED,IQ)
        IT = IA*IL-IR*IH
        IF(IT.GT.0) THEN
          ISEED = IT
        ELSE
          ISEED = IC+IT
        END IF
        RANF = ISEED/FLOAT(IC)
      RETURN
      END

PR #72, opened 2018-07-27
disambiguates nullary statement function

jasonqxu:
While we were using fortran-src to do data flow analysis on our internal code, we have seen similar errors as reported in https://github.com/camfort/fortran-src/issues/43

Currently fortran-src treats statement function with no argument as an assignment with its LHS being a function call. So code like
```
func0 () = 42
```
is parsed as
```
StExpressionAssign ()
                                                     (4:7)-(4:19)
                                                     (ExpFunctionCall ()
                                                                      (4:7)-(4:14)
                                                                      (ExpValue ()
                                                                                (4:7)-(4:11)
                                                                                (ValVariable "func0"))
                                                                      Nothing)
                                                     (ExpValue ()
                                                               (4:18)-(4:19)
                                                               (ValInteger "42"))
```


jasonqxu:
The 'disambiguateFunctionStatements' function correctly resolves function statements with at least one argument. 
The issue with nullary function statement is that it is initially parsed as 'ExpFunctionCall', not 'ExpSubscript' which is handled by 'disambiguateFunctionStatements'

mrd:
Thanks. I will go ahead and modify the typechecker so that it also recognises this situation as well.

PR #77, opened 2018-08-09
Fix exception for array declaration in common statement

jasonqxu:
For the following code which declares an array in the common statement
```
integer a
common /c/ a(10)
```
fortran-src throws exception:
```
uncaught exception: ErrorCall (Use of srcName on non-variable.
       CallStack (from HasCallStack):
         error, called at src/Language/Fortran/Analysis.hs:166:72 in fortran-src-0.2.2.0-3De5L0hl9pcKK2n6AkwMyF:Language.Fortran.Analysis)
```


jasonqxu:
Array declarator can occur in common statement as per the Fortran 77 standard http://www.fortran.com/F77_std/rjcnf-5.html#sh-5.1.2.1

mrd:
Thanks, good catch.

Issue #79, opened 2018-08-21
Add a reference to file containing token in Language.Fortran.Utils.Position


vaibhav-y:
I have a suggestion for changing `Language.Fortran.Utils.Position`, so that it includes a reference to the file. This would be useful when processing `StInclude`s so that notable events from within the included file are reported with the correct file path.

My suggestion is to have map of file to file-ids, that is passed updated with seen files while building the AST, and `Position` to hold the file id corresponding to the file the token was seen in. 

I think that having the file path as a part of the error would be beneficial for diagnostics reporting.

What are your thoughts on this?

madgen:
This seems to be useful only for the includes, which is a good enough justification now that we have some support for them.

I don't see why we need a map from files to file-ids though. Is there a case in which the file belonging to an id needs to be updated throughout the AST?

@mrd thoughts?

vaibhav-y:
The file ids is primarily for efficiency sake, storing a integer in each AST node instead of the full string which remains fixed throughout the run. As for files updated through the AST, I think it could be convenient for source to source transformations

cc: @traveltissues @ruoso

mrd:
Such strings would likely be shared structure in Haskell since they all come from the same place.

mrd:
Closed in #80.

PR #80, opened 2018-08-27
Add `filePath` field to `Position` record

vaibhav-y:
PR summary:

Related issue: #79 
```
    * Modify position record to now contain a `filePath` field

    * Update tests and associated functions relying on Position
```

cc: @mrd @madgen @ruoso @burz @traveltissues


traveltissues:
lgtm

mrd:
I need to fix camfort for this patch as well

mrd:
CamFort fixed, pull request merged but reverted change to the Show instance that printed out filename as part of Position since that broke too much. (Perhaps we should reconsider relying on the Show instance for this).

PR #81, opened 2018-08-29
Fixes

### LINKED ISSUE #76: Orphan instances remaining in source
Modules such as Language.Fortran.AST have orphans of `Spanned` and `FortranVersion`. Since these are defined in this same package I think it's reasonable to  `{-# OPTIONS_GHC -fno-warn-orphans #-}` in that module (and others which only have 'native' orphans). In cases like `instance (Typeable a, Typeable b) => Data (Gr a b) where` in Language.Fortran.Analysis what is preferred? Implementing a wrapper type is the accepted workaround.

traveltissues:
#76 

I eventually concluded that 'fixing' these last cases of orphans and duplication would have introduced some pretty horrible things into the code and that it was actually better not to fix them: Duplication in the test specs is fine, there's no rebinding, it just has some explicit steps that potentially could be marshalled into a monad but really doesn't provide any benefit. The remaining orphans are mostly package natives and in either case typename abuse doesn't really offer anything aside from fooling the compiler into not warning about them since it still won't provide type safety

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ script:
 - echo $TRAVIS_TAG
 - "GHC_OPTIONS=\"--ghc-options=\"$(ruby -e 'print (ENV[%q|TRAVIS_TAG|] =~ /v/ ? %q|-O3| : %q|-O0|)')\"\""
 - echo $GHC_OPTIONS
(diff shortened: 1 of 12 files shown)


traveltissues:
@mrd , what's your opinion on these changes? I admit that most of them are just encoding acknowledgement of potential issues.

traveltissues:
@mrd does this look ok now?

mrd:
Thanks

PR #83, opened 2018-09-04
Add analyseBBlocksWithoutRewritingFunctionCall in BBlock

jasonqxu:
Function `analyseBBlocks` in the `BBlock` module rewrites function calls to model Fortran's call-by-reference semantics. In our use case, we need to match function calls after data flow analysis. The rewriting makes the matching difficult as the client code of  `analyseBBlocks` does not know how the function call has been changed. `analyseBBlocksWithoutRewritingFunctionCall` does the same as `analyseBBlocks` except keeps the function calls intact in the AST.

### DIFF
--- a/src/Language/Fortran/Analysis/BBlocks.hs
+++ b/src/Language/Fortran/Analysis/BBlocks.hs
@@ -2,7 +2,7 @@
 
 {-# LANGUAGE FlexibleContexts, PatternGuards, ScopedTypeVariables #-}
 module Language.Fortran.Analysis.BBlocks
-  ( analyseBBlocks, genBBlockMap, showBBGr, showAnalysedBBGr, showBBlocks, bbgrToDOT, BBlockMap
+  ( analyseBBlocks, analyseBBlocksWithoutRewritingFunctionCall, genBBlockMap, showBBGr, showAnalysedBBGr, showBBlocks, bbgrToDOT, BBlockMap
   , genSuperBBGr, SuperBBGr, showSuperBBGr, superBBGrToDOT, superBBGrGraph, superBBGrClusters, superBBGrEntries
   , findLabeledBBlock )
 where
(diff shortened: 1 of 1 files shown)


ruoso:
> It looks like analyseBBlocksWithoutRewritingFunctionCall would break interprocedural dataflow analysis. 

I'm not sure I understand why the rewriting is needed for the interprocedural dataflow analysis...

ruoso:
> Do you need basic blocks at all or can you stick to the AST? 

We are doing intra-procedure data flow analysis for the purposes of rewriting the code, so we do take benefit from the basic blocks, we do have encoded the knowledge that a subroutine or function call takes those values by reference and may change its values, but for us it's more important to be able to more easily match with the original AST and source code.

mrd:
A basic block is a series of instructions with straight-line control-flow, e.g. no branches. The final instruction in each basic block should be some form of branch (e.g. StCall, StGoto, or the implicit branch after an the body of an if-block, etc). A function call is a branch (and return branch). The basic blocks analysis bottom-up recursively rewrites all function call expressions into subroutine-style StCalls with the arguments and returns copied into temporary variables. The interprocedural analysis works by stitching together the basic block graphs of the different functions and subroutines; it pattern-matches on the StCall instruction found at the end of basic blocks in order to obtain the corresponding basic block graph.

ruoso:
> A function call is a branch (and return branch). The basic blocks analysis bottom-up recursively rewrites all function call expressions into subroutine-style StCalls with the arguments and returns copied into temporary variables.

I'm not sure I understand why a function call is fundamentally different than a subroutine call when it comes to the analysis of the execution graph, and why the temporary variables change that...

jasonqxu:
@mrd We mainly do intra-procedural analysis, so we don't need to stitch the function call with its definition, which could be defined in a separate file. So in our case, we do not assume any knowledge of the function body albeit that the arguments could be updated.

mrd:
> I'm not sure I understand why a function call is fundamentally different than a subroutine call when it comes to the analysis of the execution graph, and why the temporary variables change that...


A function call is an expression (`ExpFunctionCall`), which may appear anywhere expressions are permitted within a statement or block. They may appear multiple times, they may be nested, with the return value of one function call being fed as the argument to another function call. Therefore, to comprehensively transform these function calls into straight-line code, we recursively descend depth-first / left-to-right and replace all function calls with temporary variables while outputting straight-line code to set up the arguments for the sequence of `StCall` statements that replace the function calls. Furthermore, the return value of a function call is implicitly rewritten as the first argument of the generated `StCall`.

Example: `x = f(g(y),h(z))` becomes something like:

`t1 = y`
`StCall g(t2, t1)`
`t3 = z`
`StCall h(t4, t3)`
`StCall f(t5, t2, t4)`
`x = t5`

With this straight-line code the basic-blocker can work simply by cutting blocks that end after each `StCall`. This effectively mimics what a typical compiler does when converting AST into internal representation on its way to producing assembly language output.

mrd:
> Being able to defer that linearization is very important for us because we want to be able to evaluate the data flow as closely as possible to the original source code, such that we can drive localized code rewriting.

It's possible to do dataflow analysis without flattening or linearisation of the code. However the algorithm that we use makes the assumption that control-flow is straight-line through the basic block and I would be very uncertain of its behaviour with `ExpFunctionCall` present still. Certainly it has not been accounted for in helper functions that figure out uses and defs.

I think this may need to be in a different module, as you are describing a different form of internal representation. I admit that part of the reason you've been able to reach this point is my own laziness (a Haskell programmer being lazy!). I should have designed a new algebraic datatype for the basic block intermediate representation, one that could not admit nested expressions at all. That would be the typical approach for a compiler, e.g. using a three-address code. However, Fortran is such a procedural language that a subset of our AST datatype was already available and easily reusable for the basic block intermediate representation.

ruoso:
> I think this may need to be in a different module, as you are describing a different form of internal representation.

Fair enough, do you think it would be possible to factor out the statement-level graph as a separate function that gets used by both cases?

mrd:
Possibly? If it doesn't make things too unreadable. I take it from your description that you wouldn't want `StCall` to terminate these 'not-so-basic' blocks, also, since for intra-procedural analysis they don't alter control-flow.

I won't be able to help much over the next two weeks, as I'm sporadically available. Sorry about that.

jasonqxu:
> Do you need basic blocks at all or can you stick to the AST?
What we really need is the control flow graph, the node of the CFG does not need to be basic block, actually it would be more convenient for our use case if the node is at the statement level.

jasonqxu:
@mrd, Thanks for your input on the PR. We are going to pursue an alternative solution for the problem of  matching function calls after  `analyseBBlocks` of AST .

PR #84, opened 2018-10-02
Format src/ using brittany defaults

traveltissues:
We have been applying formatting standards to various projects to increase legibility. in order to reduce synchronisation friction with our fork it would be more convenient to apply those here. This makes use of the default style enforced by [brittany](https://hackage.haskell.org/package/brittany) but I'm not committed to those defaults.


mrd:
It's an interesting idea. I'm not keen on some of the choices brittany has made, especially regarding alignment. I've been digging through his website and documentation and unfortunately I can't find much information about custom configuration options. It seems to be an outstanding problem. Do you have some better info?

traveltissues:
the config info is not well documented at the moment but configs are yaml and by default it's
```
conf_debug:
  dconf_roundtrip_exactprint_only: false
  dconf_dump_bridoc_simpl_par: false
  dconf_dump_ast_unknown: false
  dconf_dump_bridoc_simpl_floating: false
  dconf_dump_config: false
  dconf_dump_bridoc_raw: false
  dconf_dump_bridoc_final: false
  dconf_dump_bridoc_simpl_alt: false
  dconf_dump_bridoc_simpl_indent: false
  dconf_dump_annotations: false
  dconf_dump_bridoc_simpl_columns: false
  dconf_dump_ast_full: false
conf_forward:
  options_ghc: []
conf_errorHandling:
  econf_ExactPrintFallback: ExactPrintFallbackModeInline
  econf_Werror: false
  econf_omit_output_valid_check: false
  econf_produceOutputOnErrors: false
conf_preprocessor:
  ppconf_CPPMode: CPPModeAbort
  ppconf_hackAroundIncludes: false
conf_version: 1
conf_layout:
  lconfig_altChooser:
    tag: AltChooserBoundedSearch
    contents: 3
  lconfig_importColumn: 60
  lconfig_hangingTypeSignature: false
  lconfig_alignmentLimit: 30
  lconfig_indentListSpecial: true
  lconfig_indentAmount: 2
  lconfig_alignmentBreakOnMultiline: true
  lconfig_cols: 80
  lconfig_indentPolicy: IndentPolicyFree
  lconfig_indentWhereSpecial: true
  lconfig_columnAlignMode:
    tag: ColumnAlignModeMajority
    contents: 0.7
```
the internal config info is at https://github.com/lspitzner/brittany/tree/4497fa927fc77a41f49ee385af4c35f391cc5a7f/src/Language/Haskell/Brittany/Internal/Config

traveltissues:
which features would you prefer changed from the defaults? should alignments remain as they were?

mrd:
After playing with it for a while I think the key change might simply be cols: 100

Yes that's longer than traditional but 80-col layout is wrapping so many things in an ugly way.

mrd:
In any case, I won't merge this particular pull request, but instead I will add it conceptually by setting up a configuration file for Brittany and the necessary git-hooks to make it work, once I figure it out.

mrd:
I've set up Travis-CI to check that the code follows the style generated by brittany. Currently only in the brittany branch.

But I've also come across another problem, which is that the behaviour of brittany downloaded and installed from github HEAD varies significantly from most recently released version of brittany.

I could stick to the released version but this means that there will be substantial reformatting upon release of the next version.

Alternatively we could stick with the HEAD revision but that could change over time as well, quite unpredictably it seems.

It's somewhat alarming that there is such a big change in style even when the configuration remains exactly the same.

traveltissues:
yeah, that is weird. i know the tool is still in development and doesn't handle some things very well but i wouldn't have expected the style to change very fundamentally at this point. If the style changes are in opposition to the config then ofc that's a bug but otherwise it would be the result of deficient configuration options

mrd:
Yes even with the default config there is a difference between released version brittany-0.11.0.0.x2 and git HEAD.

```
--- src/Language/Fortran/Analysis.hs	2018-10-10 13:32:18.181029000 +0100
+++ /dev/fd/63	2018-10-10 13:33:00.862172266 +0100
@@ -157,17 +157,16 @@
   docPrec _ = doc
 
 analysis0 :: a -> Analysis a
-analysis0 a = Analysis
-  { prevAnnotation = a
-  , uniqueName     = Nothing
-  , sourceName     = Nothing
-  , bBlocks        = Nothing
-  , insLabel       = Nothing
-  , moduleEnv      = Nothing
-  , idType         = Nothing
-  , allLhsVarsAnn  = []
-  , constExp       = Nothing
-  }
+analysis0 a = Analysis { prevAnnotation = a
+                       , uniqueName     = Nothing
+                       , sourceName     = Nothing
+                       , bBlocks        = Nothing
+                       , insLabel       = Nothing
+                       , moduleEnv      = Nothing
+                       , idType         = Nothing
+                       , allLhsVarsAnn  = []
+                       , constExp       = Nothing
+                       }
 
 -- | True iff the expression can be used with varName or srcName
 isNamedExpression :: Expression a -> Bool

```

traveltissues:
that's no good. Also i've found that brittany seems to be a bit inconsistent with formatting of  `let` blocks. I'm going to close this for now. Downstream we'll just make this package an exception. Thanks

mrd:
We do have an interest in implementing this and I've left the scripts in the brittany branch for the time being, but it'll have to wait until this gets sorted out -- when upstream has either stabilised or fixed the bug.

PR #85, opened 2018-10-04
Disambiguate function call

jasonqxu:
In the following code `foo` is a function defined in other compilation unit. It is declared as a variable however is being used as an array in this compilation unit. With this knowledge we shall be able to resolve `foo` as a function.

```
      program main 
      integer a
      integer foo
      a = foo (1, 2)
      end
```


src/Language/Fortran/Transformation/Disambiguation/Function.hs
```diff
@@ -42,6 +42,9 @@ disambiguateFunctionCalls = modifyProgramFile (trans expression)
     expression (ExpSubscript a1 s v@(ExpValue a _ (ValIntrinsic _)) indicies)
       | Just (IDType _ (Just CTIntrinsic)) <- idType a
       , indiciesRangeFree indicies = ExpFunctionCall a1 s v (Just $ aMap fromIndex indicies)
+    expression (ExpSubscript a1 s v@(ExpValue a _ (ValVariable _)) indicies)
+      | Just (IDType _ (Just CTVariable)) <- idType a
```
traveltissues: can't you just add that as another guard for the above?

mrd:
Thanks

PR #87, opened 2018-11-28
Fix lines truncation in 'fortran77 legacy with includes' parser

lukaszkolodziejczyk:


lukaszkolodziejczyk:
@mrd 
This one-liner change introduces proper truncation of lines for 'fortran77legacy-with-includes' parser.
The other fortran77legacy parser already does that: https://github.com/camfort/fortran-src/blob/3196bc37fed5ab8577962dbb5b4b5c2549fc3fd7/src/Language/Fortran/Parser/Fortran77.y#L1114

Issue #46, opened 2017-06-08
Off-by-one error in `Util.Position.getSpan`?


buggymcbugfix:
`Util.Position.getSpan` seems to have an off-by-one error, which gets revealed in the source spans that are spat out by `units-check` as part of the constraints.

They are all short by one, e.g. `(12:9)-(12:19)` should really be `(12:9)-(12:20)`.

This is assuming that we start both lines and columns at 1 (from my observation Vim and Atom do this, but Emacs starts lines at 1 and columns at 0).

dorchard:
This can observed apart from `units-check`, e.g. take the file
```
program tiny
  implicit none
  real :: x
end program tiny
```
and putting this through `fortran-src`, we get the following AST (fragment):
```
........
     (3:11)-(3:11)
                                                       [DeclVariable ()
                                                                     (3:11)-(3:11)
                                                                     (ExpValue ()
                                                                               (3:11)-(3:11)
                                                                               (ValVariable "x"))
                                                                     Nothing
                                                                     Nothing]))]
```
Note that the span for `x` has zero length.

vimuel:
@madgen Can you give me a pointer to where I should start looking to fix this?

madgen:
@vimuel `Lexer/FreeForm.x` in particular `alexGetByte`. If you have any questions as to how it work just let me know.

harry-clarke:
@dorchard, from that example code, are all the end spans one off?
Looking at the file 1-indexed, the 11th character of line 3 is where the variable "x" starts.

There is no 12th character, but you could say that variable "x" should have span (3:11)-(3:12) if all end spans should point to the character after the end of the statement.

dorchard:
That is a good question- I guess we have to pick a convention and stick to it- I think a lot of the camfort side code assumes that the upper bound is exclusive, i.e. a single letter variable like `x` will have a span (l:c)-(l:c+1).
Is this convention used in fortran-src @madgen and @mrd ?

GuiltyDolphin:
@dorchard I agree... And we should ensure we can't violate this convention (i.e., deal with it in fortran-src and prevent access to span internals).

burz:
This is a ***big*** problem for `SrcSpan`s related to `BlIf` (i.e. if blocks).

As a trivial example, consider the following:

```fortran
      subroutine foo
      if (.TRUE.) then
      print *, 'I print'
      endif
      end
```

[...]

PR #88, opened 2018-12-19
Fix SrcSpan of "if" to include "endif"

burz:
Fixes the issue mentioned here https://github.com/camfort/fortran-src/issues/46#issuecomment-438057635

Copying here:

There is a ***big*** problem for `SrcSpan`s related to `BlIf` (i.e. if blocks).

As a trivial example, consider the following:

```fortran
      subroutine foo
      if (.TRUE.) then
      print *, 'I print'
      endif
      end
```

The `SrcSpan` for this block shows that its end is the the last character (`'`) in the `print` line.

Ok, so maybe this isn't so bad, we can *probably* assume that the real end to the `BlIf` is one additional line down and `6 + 5` characters over (6 for indentation).

On the other hand the problem is *much much* worse for if the if blocks are *nested*.

```fortran
      subroutine bar
      if (.TRUE.) then
      if (.TRUE.) then
      if (.TRUE.) then
      if (.TRUE.) then
      print *, 'I print'
      endif
      endif
      endif
      endif
      end
```

For ***every*** `BlIf` in the above code, the corresponding `SrcSpan`'s end is the last character (`'`) in the `print` line.

This means that in order to even think about determining the span of any if statement anywhere using `fortran-src` we need to traverse the tree all the way down and pull up the (possibly different) end of the span on the way up. *And this may not even be able to be done correctly if the `endif`s are not all the way to the left (i.e. 6 characters from the left).*

I have a very hard time thinking of any reasons why the current system would be useful given the nesting problem.

---

This PR fixes this issue, by changing the `SrcSpan` to include the `StEndif`, which previously was just thrown away in the grouping step.

@ruoso

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -76,3 +76,4 @@ tests:
     - deepseq
     - hspec >=2.2 && <3
     - fortran-src
+    - raw-strings-qq >= 1.1
--- a/src/Language/Fortran/Transformation/Grouping.hs
+++ b/src/Language/Fortran/Transformation/Grouping.hs
@@ -86,9 +86,9 @@ groupIf' (b:bs) = b' : bs'
     (b', bs') = case b of
       BlStatement a s label st
         | StIfThen _ _ mName _ <- st -> -- If statement
-          let ( conditions, blocks, leftOverBlocks, endLabel ) =
+          let ( conditions, blocks, leftOverBlocks, endLabel, endStmt ) =
                 decomposeIf (b:groupedBlocks)
-          in ( BlIf a (getTransSpan s blocks) label mName conditions blocks endLabel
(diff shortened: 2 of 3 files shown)


madgen:
Largely, this looks good to me (apart from the -Werror related CI failure). I don't know if adding the quasi-quoted string dependency is necessary as it doesn't seem to improve the test case. But I don't particularly mind.

Presumably this problem also happens with other kind of blocks like do blocks. Any chance you'd like to fix those as well? 😇 

As I said (after removing the redundant import), it is good for me, but it is up to @mrd to merge as he is the active maintainer.

burz:
@madgen I might be able to look into it when I find some time. I tried to do this originally but ran into some issues where one of these was not grouping its `Statement`s into a `Block` correctly. I don't recall which one it was in particular.

PR #89, opened 2019-01-02
Implicit function calls should be disambiguated

azeemba:
Currently implicit function calls show up as an array index. They
should default to be a function call when no type is declared.

### DIFF
--- a/src/Language/Fortran/Transformation/Disambiguation/Function.hs
+++ b/src/Language/Fortran/Transformation/Disambiguation/Function.hs
@@ -40,6 +40,8 @@ disambiguateFunctionCalls = modifyProgramFile (trans expression)
       , indiciesRangeFree indicies = ExpFunctionCall a1 s v (Just $ aMap fromIndex indicies)
       | Just (IDType _ (Just CTVariable)) <- idType a
       , indiciesRangeFree indicies = ExpFunctionCall a1 s v (Just $ aMap fromIndex indicies)
+      | Nothing <- idType a
+      , indiciesRangeFree indicies = ExpFunctionCall a1 s v (Just $ aMap fromIndex indicies)
     expression (ExpSubscript a1 s v@(ExpValue a _ (ValIntrinsic _)) indicies)
       | Just (IDType _ (Just CTIntrinsic)) <- idType a
       , indiciesRangeFree indicies = ExpFunctionCall a1 s v (Just $ aMap fromIndex indicies)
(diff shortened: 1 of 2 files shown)


ruoso:
>  it shouldn't be classified as a function call. Why do you think otherwise?

That's not true, tho...

```
$ cat foo.f
       subroutine foo
       integer*4 a
       a = i(10)
       end
$ gfortran-8 -c foo.f -o foo.o
$ objdump -rd foo.o

foo.o:     file format elf64-x86-64


Disassembly of section .text:

0000000000000000 <foo_>:
   0:	55                   	push   %rbp
   1:	48 89 e5             	mov    %rsp,%rbp
   4:	48 83 ec 10          	sub    $0x10,%rsp
   8:	bf 00 00 00 00       	mov    $0x0,%edi
			9: R_X86_64_32	.rodata
   d:	b8 00 00 00 00       	mov    $0x0,%eax
  12:	e8 00 00 00 00       	callq  17 <foo_+0x17>
			13: R_X86_64_PC32	i_-0x4
  17:	89 45 fc             	mov    %eax,-0x4(%rbp)
  1a:	90                   	nop
  1b:	c9                   	leaveq 
```

ruoso:
Essentially, any variable that is not declared as an array or character of kind > 1 but is used with an array subscript expression is implicitly an external function call...

madgen:
@ruoso Fair. I looked through the standard and didn't find anything that clarifies the issue. If NAG compiler under strict checks behave the same way, I don't see why not accept the change.

ruoso:
> If NAG compiler under strict checks behave the same way

Did you mean you tested in that compiler? I don't have that compiler on hand...

gfortran doesn't even raise a warning with `-Wall -Wextra`

madgen:
@ruoso yeah, @mrd should have it. I'm sure he'll have a look.

I'm weary of trusting gfortran because it basically does whatever it wants and I've seen it accept programs that are strictly prohibited by the standard and generate garbage.

ruoso:
I did find the following in the [standard](https://www.fortran.com/F77_std/f77_std.html) (section 18.2.2, External Function):

> A symbolic name is the name of  an  external  function  if  it  meets either of the following conditions:
> (1) The name appears immediately following the  word FUNCTION  in  a  FUNCTION  statement or the word ENTRY in an ENTRY statement  within  a  function subprogram.
> (2) It is not  an  array  name,  character  variable name,   statement   function   name,   intrinsic function name,  dummy  argument,  or  subroutine name,   and   every  appearance  is  immediately followed by  a  left  parenthesis  except  in  a type-statement,  in an EXTERNAL statement, or as an actual argument.

PR #95, opened 2019-03-26
Fix missing edge in StComputedGotos

burz:
As it stands in `master` at the moment, `StComputedGoto`s are missing an edge out in the produced `BBlock`s.

[According to the standard](https://www.fortran.com/F77_std/rjcnf0001-sh-11.html#sh-11.2), if the expression supplied to a computed goto does not index into the supplied list of labels then the statement should act as if it were a `continue`. However, in the current implementation the `StGotoComputed` is handled as if it were a final block with an explicit control transfer, and adds only the edges that point to the supplied labels. This means that entire swaths of code can be missing from basic blocks if the path to that code is hit only in the case of the `goto` acting as a `continue`.

The issue in practice can be seen in the supplied test case where the lines
```fortran
 10    goto (30, 40) s
 20    goto 999
```
where `10` has only 2 edges going from it in the original basic block graph produced while `20` has  none, as they are pruned out as "unreachable".

This PR fixes this issue by adding an edge to the next block in the `perBlock` handling step, while also adding the control transfer edges as before.

@ruoso


mrd:
Thanks.

PR #96, opened 2019-04-01
Add missing edges for READ in BBlocks

burz:
[According to the standard](https://docs.oracle.com/cd/E19957-01/805-4939/6j4m0vnat/index.html), the `READ` function has two optional labeled arguments, `END` and `ERR`, that allow the programmer to specify labels where control should continue when the end of file or some error occurs during the processing of the `READ`, respectively.

Currently these edges are not added into the basic blocks, which can lead to large paths being absent from the generated basic block graph.

This PR adds in the edges when these arguments are present.

@ruoso


mrd:
Thanks.

PR #97, opened 2019-04-02
Fix SrcSpan end issue for BlDo and BlDoWhile

burz:
This fixes the issue that `BlDo` and `BlDoWhile` do not have the correct end in their respective `SrcSpan`s after the grouping transformation. This is similar to how the [`BlIf` was previously fixed](https://github.com/camfort/fortran-src/pull/88).

As far as I can tell this fixes all of the post-grouping spans for FORTRAN 77 standard constructs.

Since it was rather unnecessary, I also removed the test suite's dependency on `raw-strings-qq` in favor of `unlines` as well.

@ruoso

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -76,4 +76,3 @@ tests:
     - deepseq
     - hspec >=2.2 && <3
     - fortran-src
-    - raw-strings-qq >= 1.1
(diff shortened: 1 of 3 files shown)


mrd:
Thanks.

PR #98, opened 2019-04-04
Handle leading-zero labels in BBlocks

burz:
FORTRAN considers labels with leading zeroes to be the same as those without them, for instance the following program still has a valid `goto`
```fortran
      program foobar
      goto 400
      print *, 'unreachable'
      goto 900
 0400 print *, '0400'
  900 continue
      end
```
However, `fortran-src` is not picking up on this in `BBlocks`, so in the preceding example the basic block containing `0400` would end up being unreachable.

This PR fixes this issue by simply removing the leading zeroes from the strings used as keys in the label map. It is done this way as opposed to simply converting the strings to their numerical representation because it seems the label map also contains variables as keys in the case of the F77 `ASSIGN` statement. Since variables cannot start with numbers, this should not affect this code path.

@ruoso


mrd:
Thanks.

PR #106, opened 2019-08-19
Parse nested structure

RaoulHC:
Despite the fact that `StructStructures` was a valid `StructureItem` in the AST, there were no rules for parsing it.

I've added this to `fortran77.y`, as well as changing `StructureItem` to have a necessary name for accessing the substructure. This isn't the full possible syntax of "field-lists" for naming these nested structures, but this could be addressed in a future PR  (for sun see [here](https://docs.oracle.com/cd/E19957-01/805-4939/6j4m0vn6t/index.html), for xlf see record structures in chapter 21 [here](https://www-01.ibm.com/support/docview.wss?uid=swg27024776&aid=1), and for gfortran see [here](https://gcc.gnu.org/onlinedocs/gfortran/STRUCTURE-and-RECORD.html))


mrd:
Thanks, can go into next version.

PR #107, opened 2019-08-29
Ignore comments in structure declaration

jasonqxu:
The `Fortran77Legacy` parser supports parsing of the structure, union and map. The following strcuture is parsed as  a single `StStructure` node. 
```fortran
structure /foo/
  integer i
  union 
    map
      integer j
    end map
    map
      real r
   end map
 end union
structure
```
Comments within the structure declaration are tricky to be included in the AST so they shall just be ignored.
Fortran77Legacy lexer used to ignore all comments. Now It passes the comment tokens to the parser for use cases specified in the PR
https://github.com/camfort/fortran-src/pull/103

However, the parsing of structure now fails if there is any comment inside the structure declaration. This PR update the grammar for structure, union, map to ignore the comments within the structure declaration.


mrd:
Thanks.

PR #117, opened 2020-01-22
Add subscript to ARITHMETIC_CONSTANT_EXPRESSION

RaoulHC:
Fortran intrinsics are allowed in these types of expressions, so the parser needs to be able to parse functions here. For example in setting the length of a character variable, the following is valid: `character*(len(d)) c`.

There might be a better solution as this allows a lot of non-valid expressions, but given function calls need to be disambiguated from other subscript expressions maybe this isn't possible in the parser.


mrd:
Ok, we'll see how it goes. Thanks.

PR #119, opened 2020-07-07
Pretty printer fixes

RaoulHC:
Two small fixes to the pretty printer which resulted in either failures or invalid code.
* Avoid parenthesis for integer literals in the kind and length selectors, for example GFortran wont compile the declaration `integer*(4) i`.
* Pretty print old style initialisers. Previously it failed saying that variable initialisation wasn't in fortran 77, despite the fact the parser accepts it.


mrd:
Thanks

PR #120, opened 2020-07-27
Update lts

RaoulHC:
We're hoping to use ghc 8.8 soon so this moves the fail definition out of Monad into MonadFail, as well as updating various pedantic warnings introduced by changes to libraries.


mrd:
thanks

PR #121, opened 2020-07-27
Don't truncate lines after 72 but skip in the lexer

RaoulHC:
Previously fortran-src just truncated lines for legacy 77 mode before any lexing, and this meant `PosAbsoluteOffset` didn't correspond to the position in the file for anything following characters after column 72.


mrd:
thanks

PR #122, opened 2021-01-04
Promote FortranVersion to own separate module

raehik:
`FortranVersion` is used by various parts of the project and by fortran-src users, so placed into its own top-level module.

For now, `ParserMonad` continues to export it in order to keep API from breaking. That is, `import Language.Fortran.ParserMonad ( FortranVersion(..) )` will continue to work, but should probably be consider deprecated.

### DIFF
--- a/package.yaml
+++ b/package.yaml
@@ -42,6 +42,7 @@ library:
   - Language.Fortran.Analysis.BBlocks
   - Language.Fortran.Analysis.DataFlow
   - Language.Fortran.AST
+  - Language.Fortran.Version
   - Language.Fortran.LValue
   - Language.Fortran.Intrinsics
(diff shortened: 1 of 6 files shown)


raehik:
Looks like the `deduceVersion` change (move+rename) has some breaking changes: CamFort and fortran-common each use it once. I don't think it belongs in Parser.Any, so I'd push to keep the change and update those projects. Also means we can update the other version imports, then finish off with deprecating `import ParserMonad ( FortranVersion(..) )`.

raehik:
Half related, I'm going to take a look at the Travis config. Looks like the config is trying to use too old packages.

raehik:
Weird rebase bug (?), didn't update from master correctly. Force-pushed and solved.

PR #123, opened 2021-01-05
Fix Travis build; small cleanup

raehik:
Added Stack lock file (appears to be good practice) and updated `.gitignore` since it was complaining about the generated `*.cabal` which shouldn't be added.


dorchard:
Should this allow the travis CI build to work on this pull request?

raehik:
Looks like Travis acts on every PR or every branch push, so yep. Travis has been failing builds for a while though due to old repos. This change gets us the latest Stack (Ubuntu repos have it, but only 1.9, and it looks like resolver YAMLs have moved since then), and then the rest still works fine.

I removed a deprecated Stack arg, so we're targeting Stack 2 a bit more clearly now. I'll add that to the README.

raehik:
Also, the build is taking just 3m 15s with caching! That's a lot faster than I remember us discussing ("~30 minutes on a standard machine").

PR #124, opened 2021-01-08
CI via GitHub Actions

raehik:
The old, free Travis CI site travis-ci.org is [closing down](https://mailchi.mp/3d439eeb1098/travis-ciorg-is-moving-to-travis-cicom) and merging with travis-ci.com. Actually, it should have happened last week, so we'll want to migrate CI soon.

However, they're also changing the free plan, and I've only heard [negative things](https://travis-ci.community/t/org-com-migration-unexpectedly-comes-with-a-plan-change-for-oss-what-exactly-is-the-new-deal/10567). So I'm trying out GitHub Actions instead.

### DIFF
--- a/.github/ci.yml
+++ b/.github/ci.yml
@@ -1,0 +1,32 @@
+name: CI
+
+on:
+  push:
+    branches:
+    - main
+  pull_request:
+    types:
+    - opened
+    - synchronized
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


raehik:
Can't find a reference in GitHub's docs, but [looks like we need to have a workflow in `master`](https://github.community/t/workflow-files-only-picked-up-from-master/16129/7) before I can mess around further. So merging!

PR #125, opened 2021-01-08
CI: correct filename for GH Actions workflow

raehik:


raehik:
hah, just got the filename wrong :facepalm:
This will now be the CI playing branch.

raehik:
Ah nope, it's not letting me re-run with a workflow outside of master. Looks a bit poorly defined. Merging this, sorry for the messy history -- I'm at least rebasing...

PR #127, opened 2021-01-08
use --pedantic when building with Stack

raehik:


raehik:
OK, we have a basic GitHub Actions CI workflow implemented now alongside Travis (until the latter goes down). Any further CI changes I'll probably test on a fork first.

Issue #116, opened 2020-01-10
Reprinter does not detect if it generate very long lines


FulgurIgor:
I have tried to reprint the following code:
```fortran
      program main
      integer, parameter :: &
      long_variable_name_10 = 1, &
      long_variable_name_11 = 1, &
      long_variable_name_12 = 1, &
      long_variable_name_13 = 1, &
      long_variable_name_14 = 1, &
      long_variable_name_15 = 1, &
      long_variable_name_16 = 1, &
      long_variable_name_17 = 1, &
      long_variable_name_18 = 1, &
      long_variable_name_19 = 1, &
      long_variable_name_20 = 1, &
      long_variable_name_21 = 1, &
      long_variable_name_22 = 1, &
      long_variable_name_23 = 1, &
      long_variable_name_24 = 1, &
      long_variable_name_25 = 1, &
      long_variable_name_26 = 1, &
      long_variable_name_27 = 1, &
      long_variable_name_28 = 1, &
      long_variable_name_29 = 1, &
      long_variable_name_30 = 1, &
      long_variable_name_31 = 1, &
      long_variable_name_32 = 1, &
      long_variable_name_33 = 1, &
      long_variable_name_34 = 1, &
      long_variable_name_35 = 1, &
      long_variable_name_36 = 1, &
      long_variable_name_37 = 1, &
      long_variable_name_38 = 1, &
      long_variable_name_39 = 1
      end
```

And it generates the following output, although the maximum length of the line is 132 characters:
```fortran
program main
[...]

PR #129, opened 2021-01-26
Add continuation reformatter

raehik:


raehik:
Addresses #116 . This would be tough if the continuation feature wasn't very low-level; appears that strings can have continuations too, so no apparent need to lex. Should be tested on lots of code, though.

Think this might wreck comments and cause invalid output currently -- might need some to track whether we're in a comment.

raehik:
Problems:

  * If `!` is allowed in strings, we can't blindly parse for `!` to determine if we're in a comment - we need string lexing
  * Comments after continuations must be ignored. Currently we could e.g. insert a continuation in the comment after a continuation. We need to lex for `&` and turn off reformatting until a newline.

raehik:
Not the ideal solution, but does the job. Placed behind a default-off flag `--split-long`. Tries to work for F77 and F90.

One thought -- `c` in the first column moves us to comment mode (= don't continuate) for the line. I don't think F90 is indentation sensitive (enforced, anyway), so this could result in us not continuating where we should do. But it's 1) quite specific, 2) won't *add* any further errors, and 3) probably not ever produced by our indenting pretty printer.

raehik:
History cleaned up, good for a review.

raehik:
Resolves #129.

PR #131, opened 2021-02-04
Define some common utils in a separate module

raehik:


raehik:
All non-breaking changes. `Language.Fortran.Util.Files` now provides some generally handy functions for using with fortran-src.

PR #132, opened 2021-02-09
Migrate existing rewriter tests

raehik:
The rewriter tests from fortran-common.


raehik:
(I wanted to clean this up so it doesn't generate files during testing, but it turned out not to be an easy change since the rewriter is closely tied to the filesystem.)

PR #133, opened 2021-02-17
rewriter: temporary files via a temp sys dir

raehik:
Temporary file names when applying rewrite replacements are no longer
deterministic, so the respective test has been removed.

### DIFF
--- a/fortran-src.cabal
+++ b/fortran-src.cabal
@@ -4,7 +4,7 @@ cabal-version: 1.12
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: 2d18bd343f3cdb0783b1a174c15e83a2e3c26b2dcb261dde4e72fb3eb5a4946f
+-- hash: 48e669c2672cd1dc959b18b2b9b123b9372dfeafe821647b66bab38a32c3f8cb
 
 name:           fortran-src
 version:        0.4.1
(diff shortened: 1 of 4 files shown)


raehik:
~I really can't figure out a solution I'm happy with though. Using the system temporary directory can result in errors due to renaming across filesystems (! how is this not covered by `renameFile`??), but~ actually we can more or less always assume the directory the source file is in it writeable, so we can use that.

raehik:
urgh, none of my solutions are improvements on the existing method. The tests expect certain IO behaviour in the face of an error, which I can't seem to keep.

fortran-src.cabal
```diff
@@ -0,0 +1,161 @@
+cabal-version: 1.12
```
mrd, line 1: did you mean to commit a cabal file?

raehik: Yes - I had been keeping them out before, but I changed my mind due to warnings (when using Git repos directly in Stack `extra-deps`) and glances at other packages doing it differently. I never grasped the general consensus on it - what do you usually do?

mrd: Seems to be the recommended practice now. https://www.fpcomplete.com/blog/storing-generated-cabal-files/

PR #135, opened 2021-02-23
small tweaks

raehik:


raehik:
this doesn't matter, I've now made the spelling change elsewhere.

PR #136, opened 2021-02-25
CI: build via Cabal as well

raehik:
I don't know how to use Stack to easily test multiple GHC versions and configurations, so I bit the bullet and figured out Cabal.

Currently building Stack, plus the latest 3 GHCs (8.6, 8.8, 8.10).


raehik:
One inconvenience brought by this: now the `*.cabal` file actually matters, so should be committed whenever it changes. Though that was kinda already the case when we added it to the repo based on a recommendation.

PR #138, opened 2021-03-09
Allow selection of post-parse AST transformations

raehik:
Centralize default transformations and export more parsers (currently for Fortran 66 only).

### DIFF
--- a/src/Language/Fortran/Analysis/ModGraph.hs
+++ b/src/Language/Fortran/Analysis/ModGraph.hs
@@ -86,7 +86,7 @@ genModGraph mversion includeDirs paths = do
       iter path = do
         contents <- liftIO $ flexReadFile path
         let version = fromMaybe (deduceFortranVersion path) mversion
-        let (Just parserF0) = lookup version parserWithModFilesVersions
+        let parserF0 = parserWithModFilesVersions version
         let parserF m b s = fromRight (parserF0 m b s)
         fileMods <- liftIO $ decodeModFiles includeDirs
         let mods = map snd fileMods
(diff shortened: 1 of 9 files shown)


raehik:
@dorchard There's some documentation and work on the other parsers remaining, but the gist of the change can be seen in the [Fortran 66 parser](https://github.com/camfort/fortran-src/pull/138/files#diff-5bf082a1f3dc617089a2bf85ba63dded214f4cd0bc3b5eacb457625197a4684dR512).

Adding in 3 transformation options onto the 2 modfile options gives a total of 6 choices for parsers, which I think is the limit for what we want to export. (I'll make it clear in the docs that `xNoTrans` and `x` are aliases for `xWithTrans []` and `xWithTrans defaultTransformations`.) Is this the right direction?

dorchard:
This looks to me. (as discussed with @raehik perhaps s/Trans/Transform for clarity) if @mrd  is happy then we can apply this idea to all the other parsers.

mrd:
> This looks to me. (as discussed with @raehik perhaps s/Trans/Transform for clarity) if @mrd is happy then we can apply this idea to all the other parsers.

Fine by me - maybe unnecessary to have *NoTrans functions as the use case is limited and easily substituted.

raehik:
Agreed on removing the `*NoTransforms` functions, the derivation should be obvious for any end user.

New parsers are named with the scheme `*WithTransforms`, because using singular there feels inappropriate for a list.

I did the association list -> function refactors I mentioned, it makes an error case (no parser for given Fortran version) a lot clearer. @mrd @dorchard thoughts?

dorchard:
Looking good!

raehik:
Added a Haddock note on parser naming so the new functions won't get lost -- merging now

PR #139, opened 2021-03-15
Add static statement to legacy parser and AST

RaoulHC:
Pretty much identical to automatic statement equivalent.

I guess we never encountered it before because it's the default behaviour for legacy fortran, but it has popped up now.


raehik:
Looks good, thanks. I'll defer to @dorchard for the final word (so someone else knows)

dorchard:
Looks good to me.

PR #140, opened 2021-03-25
Replace BaseType in IDType with a semantic type representation

raehik:
Replaces `BaseType` in `IDType` with a type representation adapted from fortran-vars. Type analysis now derives `SemType` values for expressions, which include kind information and turn `DOUBLE PRECISION`s into `REAL(8)`s etc.

## Overview of changes
  * use `SemType` instead of `BaseType` in `IDType` (type analysis)
    * Constructors match original fortran-vars naming.
  * move `CharacterLen` from parsing to type analysis
    * `BaseType` values are now simple tags.
    * deleted some CharacterLen code due to no longer being in AST
  * F90 real literal parser (parse kind info)
  * export some infer monad utils (potentially useful for running just parts of type analysis)
  * some parsing fixes (type kind selectors)
  * tweaks

### DIFF
--- a/fortran-src.cabal
+++ b/fortran-src.cabal
@@ -4,7 +4,7 @@ cabal-version: 1.12
 --
 -- see: https://github.com/sol/hpack
 --
--- hash: 3eb3f3e8de907d8c4626c9351665c46eb008759e3b74f7459007de21a166fa86
+-- hash: 94796d02b259851a845bd0e2d43eae13767addbada1ac402a731b4056fe31c92
 
 name:           fortran-src
 version:        0.4.3
@@ -28,7 +28,7 @@ source-repository head
 
 library
   exposed-modules:
-      Language.Fortran.Vars.Types
+      Language.Fortran.Analysis.SemanticTypes
       Language.Fortran.Analysis
       Language.Fortran.Analysis.Renaming
       Language.Fortran.Analysis.ModGraph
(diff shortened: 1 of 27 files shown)


raehik:
Don't know where to put this so I'll upload what I have here: short program testing gfortran and fortran-src behaviour on real literals and parameters (this changeset adds parsing for the rest of the info in real literals). Behaviour is looking correct so far.

[real-literals.f90.txt](https://github.com/camfort/fortran-src/files/6326235/real-literals.f90.txt)

raehik:
There's new functionality I need to write tests for (fancier literal parsing, expected type errors/recovery behaviour). Current tests are changed to work with `SemType`.

raehik:
Copying an insight I wrote in #149 :

  > This changeset is a big step to making two distinct "type" stages -- the first having little/no post-parse processing, the second based off declaration/statement/expression analysis. I believe `TypeCharacter` stored len+kind because it's especially common to want to inspect string type metadata (length). `SemType` maintains that data for CHARACTERs, and handles the other types too.

raehik:
I think this is ready for review now. I'll use it to prepare a changeset for fortran-vars so it depends on our (slightly different) type representation.

This is the first big systemic change I've authored on camfort projects -- I don't know how exactly we should review/merge. For now, I'll write up a changelog at the top.

PR #142, opened 2021-04-06
add test case for troublesome continuation

raehik:
This seems to have different behaviour on different systems.
Want to see what CI thinks of it.


raehik:
The solution for this is still pending (I believe it would involve non-trivial lexing work), but this was a temp thing that we understand better now, so closing.

PR #146, opened 2021-04-28
see how happy CI is compiling with new GHCs

raehik:
Easier to use the machine than to Cabal my poor laptop to death.


raehik:
No problems, so I'll idly add GHC 9.0 to the versions to test. The updated pattern matching checking catches 2 redundant patterns, which is cool. (Won't bother touching them for now, don't know how the older GHCs will react.)

GHC 9.2 isn't supported by the GHC helper we're using, but I think it should be, so I've filed an issue. If we get GHC 9.2 testing too, I'll go round the other CamFort projects pushing similar CI updates. Sitting on it until then.

PR #147, opened 2021-05-11
Haddock annotate AST module

raehik:
## Changes
  * promote `AList` to module
  * Haddock TOC/ordering
  * move `SpannedPair` and generic `Spanned`, `SpannedPair` instances to `Util.Position`
  * tweaks (move instances for clarity/to remove orphans)

### DIFF
--- a/fortran-src.cabal
+++ b/fortran-src.cabal
@@ -4,7 +4,7 @@ cabal-version: 1.12
 --
 -- see: https://github.com/sol/hpack
 --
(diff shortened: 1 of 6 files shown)


raehik:
Above issue is now tracked at https://github.com/camfort/fortran-src/issues/148 , so that GHC 8.8 failure is irrelevant.

raehik:
Had a short discussion with dorchard on the important bits here, but I'm marking as draft again while there are todo tags left.

raehik:
The todos were for code that has moved in #140 , so not fussed. Merging.

PR #149, opened 2021-05-11
Remove Maybe length, kind fields in TypeCharacter

raehik:
The CHARACTER type previously stored two Maybes for length and kind. These
weren't parsed from the syntax, instead retrieved from the Selector in the
declaration's TypeSpec. With the change to type representation (past the parsing
phase), that info now goes unused.  Removing it makes BaseType more consistent
(now a "type tag" enum) and lets us remove some code.

The info has essentially been "moved" to STyCharacter in SemType (kind isn't there yet, but I plan to add it).

### DIFF
--- a/src/Language/Fortran/AST.hs
+++ b/src/Language/Fortran/AST.hs
@@ -89,7 +89,7 @@ data BaseType =
   | TypeComplex
   | TypeDoubleComplex
   | TypeLogical
-  | TypeCharacter (Maybe CharacterLen) (Maybe String) -- ^ len and kind, if specified
+  | TypeCharacter
   | TypeCustom String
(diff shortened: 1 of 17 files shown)


raehik:
This changeset is the a big step to making two distinct "type" stages -- the first having little/no post-parse processing, the second based off declaration/statement/expression analysis. I believe `TypeCharacter` stored len+kind because it's especially common to want to inspect string type metadata (length). `SemType` maintains that data for CHARACTERs, and handles the other types too.

raehik:
OK, ignore this PR -- it was useful to review separate from the other semtype changes. I've had an OK so I'm merging it into its big brother base #140 .

PR #150, opened 2021-05-13
Don't lex `;` in first 6 cols in fixed form (#148)

### LINKED ISSUE #148: Fixed form lexer can return a TInt in the first 6 columns
Detected by QuickCheck during a CI run: https://github.com/camfort/fortran-src/runs/2556618061?check_suite_focus=true#step:8:945

The fixed form lexer returns an unexpected token on `;1`. In GHCi, `safeLex66 ";1"` returns `Just (TInt 1)` instead of `Nothing` or a label/comment.

This certainly seems incorrect, but I'm not sure what the spec would say here.

raehik:
Fixes #148 .

### DIFF
--- a/src/Language/Fortran/Lexer/FixedForm.x
+++ b/src/Language/Fortran/Lexer/FixedForm.x
@@ -85,7 +85,8 @@ tokens :-
 
   <0,st,keyword,iif,assn,doo> \n              { resetPar >> toSC 0 >> addSpan TNewline }
   <0,st,keyword,iif,assn,doo> \r              ;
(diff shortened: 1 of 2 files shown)


raehik:
Upon cross-referencing some specs, the semicolon is often supported as a *statement separator* in extensions to fixed form source (gfortran, HP's F90 reference, the F90 handbook). None of them explicitly detail behaviour in this case -- but I think it's implicit from the name. Statements don't happen in the first 6 columns (it's primarily for statement labels).

This was found due to a QuickCheck property from @madgen striking gold after 5 years! cool, nice work :)

PR #151, opened 2021-05-18
Retain parsed includes to avoid reparsing them

RaoulHC:
This was causing parsing to take significantly more time and more memory
than was necessary, and there are certain files where this makes orders
of magnitude difference.


raehik:
Apologies for leaving this. Was about to suggest bringing it to the other parsers too, but was then surprised to see we don't support includes in later parsers (though I understand it's a wonky feature and >F90 has modules instead).

Thanks for this, looks good to me. I will double-check with @dorchard & see if I have his blessings to review/merge work without deferring every time :)

raehik (approved):

raehik:
Merging. I'll do a release for all these recent additions!

PR #154, opened 2021-06-11
Context free parsing

RaoulHC:
A number of constructs that were done via post parsing grouping transformations, but they do appear to be context free (labelled do's being the exception). The grouping transformations are fairly intensive as they currently need several traversals of the AST, so it's a reasonable improvement if it's done during parsing (~10-20% from a quick look).

Happy to be proved wrong if there's something I've missed that does make these other constructs context dependent!
If that's the case we might want to think about make it optional, as the vast majority of cases should be parsable like this and users might want to remove these edge cases. We may want to make this the case for the labelled do's.

If people are happy with this I'll update this to delete some of the grouping transformations, and add block parsing for select cases as well, which also appears to be context free.


src/Language/Fortran/Parser/Fortran90.y
```diff
@@ -308,6 +312,94 @@ BLOCK :: { Block A0 }
   { BlInterface () (getTransSpan $1 $8) $2 False [ ] (reverse $5) }
 | COMMENT_BLOCK { $1 }
 
+IF_BLOCK :: { Block A0 }
+IF_BLOCK
+: if '(' EXPRESSION ')' then NEWLINE BLOCKS ELSE_BLOCKS
```
RaoulHC, line 317: As a side note, I tried to simplify this rule by incorporating `[label] [id :] if` into its own rule, but then if `StIfLogical` took precedence for some reason and it failed pars when it got to `then`. Any idea why/how to get around this?

RaoulHC:
Hmm so running this over more Fortran, it seems like while if blocks can be done context free, unlabelled do's are complicated by the fact that a labelled do block can have an end do terminal statement, which isn't uncommon, which I think makes all the do blocks context dependent.

In a context dependent parser this doesn't seem like it should be too tricky to deal with, but I guess happy cannot be used for such things.

I'll update this PR some point to support context free select case blocks, and then take advantage of the fact we can specify transformations to avoid the do block transformation where we don't need it.

RaoulHC:
Got round to updating this so is ready for review.

I've left in the case statements in the AST because as it stands the Fortran77 parser will parse the statements but not group them, so wasn't sure if it made sense to add parsing of the blocks or leave as is. They do occasionally turn up in legacy code bases so we would need them to be parsed.

I also noticed that the `groupForall` isn't actually used anywhere, should it be?

raehik:
Thanks very much for this. Some general notes follow:

  * I compared ASTs from this PR and master: looks like you've also fixed the BlCase SrcSpan calculation, since current behaviour ends the span at the line before `END SELECT`, whereas this PR ends after it (matching other multiline blocks like BlDo). Thanks!
  * I see previous behaviour was to only apply the case transformation to Fortran77Extended and beyond. Wonder if that was chosen to be slightly more flexible for pre-standardised case syntax? It'd be nice to cut `StCase`s out of the AST, if possible.
    * Attempting to use cases with the Fortran 77 parser is awkward. Only the Fortran77Legacy parser parses cases, but only Fortran77Extended was set to use case grouping. *I'm not sure how Legacy and Extended are related: the former comes after the latter in the definition, but their transformations are defined the "other way around".* Also, labelled cases aren't supported by the fixed form lexer. Messy.
  * Ran a number of tests, and I think expressivity remains the same. gfortran won't compile with mismatching labels, so no longer being able to represent that is not a concern for me.

Maybe we can look into a compromise on block-based `DO` parsing later on -- even if there are some constructions we can't handle, we could use a flag to indicate whether such constructions have been parsed, and use that to potentially skip the grouping.

RaoulHC:
> I compared ASTs from this PR and master...

Nice, didn't even notice I fixed that!

> I see previous behaviour was ...

Yeah I'm unsure what the use case for the extended parser is. We only use the legacy parser internally (in fact I believe some of our libraries are made to work only with that mode), might it make sense to make the legacy parser deal with case blocks and then just scrap the extended parser? Afaict the only difference are that the lexer doesn't deal with a bunch of constructs and it does the case block transformation.

I hadn't realised but it does look like all the compilers we use support labelled blocks, even in legacy fixed form mode. Might make sense to add this support in some other PR.

> Maybe we can look into a compromise on block-based DO parsing later on ...

Yeah I think something like that would make sense, I only encountered a couple of cases where the context free parsing  didn't work and maybe it would make sense for us to just fix these occasional cases.

raehik (commented):
I attempted cleaning up repetition in the block parsers but got the same issues as before... I think I understand why now, and it seems it would be possible with changes to overall block parsing, but gains are too minor (and still copy-paste between files). Gaaah

Like the red changes. LGTM

raehik:
On @dorchard 's suggestion, is there any chance we can get @madgen 's thoughts on this, and why grouping was originally done as a post-parse step (for `IF`s and `SELECT CASE`s in particular)? Some constructions like labelled `DO`s are clearly context sensitive, but others appear straightforward to parse. Do we lose any expressivity that matters?

raehik:
Reminder half for myself -- this adds 8 shift/reduce conflicts to the Fortran 95 parser. They aren't an error, but appear to indicate the possibility of one (e.g. whenever Raoul or I tried to refactor block starts into a reusable production, we got more S/R conflicts and actual breakages). Want to check exactly where they are and if they're malignant.

madgen:
When we first did this, our goal was to do source-to-source transformations. Eagerly grouping things don't help with that since the end markers can be left implicit and compilers are happy to accept it. Also there are if statements that take labels to signal where the `then` and `else` branches start. You don't want to do parse those as blocks in a parser or a lexer. If you look at the 66 standard, there is no notion of blocks and even 70 still supports the statement versions (perhaps even 90, I can't remember).

Grouping was introduced because we wanted to do sensible functional stuff during analysis. There's probably a cheaper formulation of blocking if performance is the only concern.

So if the priorities changed that is you no longer care about source-to-source transformations and that you don't care about old non-conforming programs, this change is probably smart!

All of this said, after a lot of blood and sweat, I came to the conclusion that one must sacrifice statelessness of the lexer to keep the parser as dumb as possible (I experimented all along the spectrum) when parsing a problematic language like Fortran. So, if you want to introduce more state to the parser going forward, you might regret it down the line due to maintainability and debugability.

RaoulHC:
Hmm I had tested labelled if blocks with GFortran and Sun Studio which refuse to compile if you label an if, but don't label the corresponding end if, but it does look like XLF allows this.

Having said that I think the block parsing still makes sense if we just ignore checking the labels match up. Afaict in the Fortran 77 standard onwards, the only if statements which are allowed are arithmetic if and logical if statements, but these are distinct and still parsed with these changes. It does refer to else if and else as statements, but they are only allowed in the context of if blocks unless I'm missing something obscure.

raehik:
Thanks very much for the insights Mistral. I had nothing to add, so wrote some benchmarking code and compared performance of this branch to head. Here's the [HEAD run](https://github.com/camfort/fortran-src/runs/3045217878) and [this branch's run](https://github.com/camfort/fortran-src/runs/3045258743). Extracting the interesting tests (full parsers, with/without transformations):

  * orig F90 def trans: 1195us
  *  new F90 def trans:  835us
  * orig F90  no trans:  670us
  *  new F90  no trans:  638us
  * orig F77 def trans: 1107us
  *  new F77 def trans: 1095us
  * orig F77  no trans:  895us
  *  new F77  no trans:  916us

Default F90 parser gets a 30% performance increase, F77 has a slight performance loss when performing no transformations. The other parsers probably follow this pattern (this tests both lexers and all transformations). They're only micro-benchmarks, not sure how this might change on larger files, uniplate traversals probably speed up some.

I'm leaning towards this more stateful approach, but I'm well aware I've not seen much Fortran in my lifetime. Will talk to Dominic and see what his opinion is.

raehik:
Note that arithmetic IFs (take labels for then/else) and logical IFs (essentially a restricted arithmetic IF) aren't touched by this changeset, nor are they grouped in a transformation currently. The Fortran 66 parser isn't touched.

We will lose the ability to represent some invalid programs. However, the transformations forbid the same, and are on by default (a recent change lets you select only the ones you want). I think the blocking is useful enough that we more or less always want it.

RaoulHC:
Updated to remove the checks for construct names given some compilers (XLF at least) are a bit more flexible with them matching. Looking at the previous grouping behaviour it seems that this was actually not allowed (see [here](https://github.com/camfort/fortran-src/blob/master/src/Language/Fortran/Transformation/Grouping.hs#L145)), but probably makes sense to allow it when some compilers do.

Also cleared up some of the unused case statement grammar which I should have removed earlier.

raehik:
Thanks again for all this Raoul, should get this merged today!

raehik:
I've noted this changeset prevents adding inline comments to various block-related statements. There's already block parsing code that handles these in a given way. It's not pressing, so I'll merge this and do the fix in another PR.

PR #156, opened 2021-06-15
Add more SemType utils

raehik:
A module in CamFort uses `Analysis` results and finishes by generating a fresh AST, so as it stands it needs to convert a `SemType` back down to a `BaseType`. It's a potentially useful utility, so I've added it here.


raehik:
I suppose it'd be best to whip up a few tests for the `TypeSpec` generation function, so I'll convert to draft until that's done. Won't do anything for Pretty instance and will leave the to-dos, because it's all somewhat internal/extra stuff anyway.

raehik:
I want to push fortran-src 0.5.0 to allow projects using fortran-src to pull from Hackage instead of a GitHub commit, so I'm merging this sneakily.

PR #157, opened 2021-07-08
start on benchmarking (parsers)

raehik:


dorchard:
Can we merge this @raehik ? It will be useful to have this benchmarking code around?

PR #159, opened 2021-07-22
Parse inline comments in if/case block statements

raehik:


raehik:
Fixes some comment parsing issues introduced by #159 - but not all, I've just found. Code blocks like these still error out:

```fortran
select case (x)
! line between
case (2)
  print *, 'foo'
end select
```

I'm not sure why -- I thought the `COMMENT_BLOCK` rule would handle them.

raehik:
Sorted and filed issue #160 .

raehik:
Behaviour is now the same as the previous parser(s) as far as I can tell.

PR #162, opened 2021-07-28
Swap DO grouping transformation order

raehik:
Fixes an issue with nonmatched END DOs, due to how groupings are
implemented.


raehik:
Note I say "swapped": this is really "swapped back" from a previous commit that introduced the problem.

PR #163, opened 2021-07-28
Uniplate direct

raehik:
(building, so may use for viewing tests)


raehik:
I've written instances for all combinations between `ProgramFile` and `Expression`. All tests pass and I've confirmed that the direct instances are being used. Now I need to check performance and go through the possibly unlawful patterns I've written. In short, there are ways to "promote" from

  * Expr->Stmt
  * Stmt->Block
  * Block->ProgramUnit

but some don't appear likely, or intended, or allowed during parsing. So I figured "lying" to Uniplate and stating that those promotions weren't possible would potentially improve performance. I want to know if there's any real gain for the more questionable lies & generally clarify behaviour.

PR #167, opened 2021-08-09
Pass the expanded filepath for includes

RaoulHC:
Previously the include expansion would find the filepath but only use
the path specified in the include statement itself, but having the
expanded path is useful for analysis.

Happy to make this optional behaviour, but I wasn't sure if there were cases where old behaviour would be preferable. Clang tooling we use also behaves along these lines (you can try `clang -Xclang -fsyntax-only -ast-dump` on some file with includes to have a look).

### DIFF
--- a/test/Language/Fortran/Parser/Fortran77/IncludeSpec.hs
+++ b/test/Language/Fortran/Parser/Fortran77/IncludeSpec.hs
@@ -1,5 +1,6 @@
 module Language.Fortran.Parser.Fortran77.IncludeSpec where
 
+import System.FilePath
 import Test.Hspec
 import TestUtil
 
@@ -23,7 +24,7 @@ spec =
                           "      include 'foo.f'",
                           "      end"
                          ]
-        incs = ["./test/Language/Fortran/Parser"]
+        inc = "./test/Language/Fortran/Parser"
         name = "bar"
         pf = ProgramFile mi77 [pu]
         puSpan = makeSrcR (6,7,1,"<unknown>") (48,9,3,"<unknown>")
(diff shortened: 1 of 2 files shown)


test/Language/Fortran/Parser/Fortran77/IncludeSpec.hs
```diff
@@ -32,10 +33,11 @@ spec =
 
         -- the expansion returns the span in the included file
         -- it should return the span at the inclusion
-        st2Span = makeSrcR (6,7,1,"foo.f") (14,15,1,"foo.f")
-        declSpan = makeSrcR (6,7,1,"foo.f") (14,15,1,"foo.f")
-        typeSpan = makeSrcR (6,7,1,"foo.f") (12,13,1,"foo.f")
-        blockSpan = makeSrcR (14,15,1,"foo.f") (14,15,1,"foo.f")
+        foo = head incs </> "foo.f"
```
raehik, line 36: perhaps bit clearer without `head` by changing `incs` to one `inc :: FilePath` and wrapping it when parsing like `iParser [inc] source`?

raehik (changes requested):
Thanks. This looks like more useful behaviour. I imagine it didn't come up before because we don't appear to give much access to parse-time include expanding on the CLI side.

To check my understanding, was the original behaviour that an include file would be searched for in a list of include directories, but the include directory it was found in wasn't kept?

RaoulHC:
Updated with that little change.

And yeah, just meant that the information was found but not used apart from just reading in the file.

raehik:
cheers!

PR #168, opened 2021-08-11
CI: build for Windows and upload Win+Lin executables

raehik:
Did this for a personal project the other day and realised it'd be nice to have here.
Not bothering with matrix building on Windows, only building on GHC 9.0.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -9,10 +9,13 @@ on:
     - opened
     - synchronize
 
+env:
+  EXE_NAME: fortran-src
+
 jobs:
-  build-test-ubuntu-stack:
+  ubuntu-stack-test:
     runs-on: ubuntu-latest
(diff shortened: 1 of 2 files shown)


raehik:
Interesting. Based on the failing tests, I assume the problem is Windows vs. Linux line endings. But I can't tell from the CI log -- even the raws -- because the failing tests are printed line-by-line, which appears to enforce Windows-style `\r\n` line endings, so they appear to match (confirmed in a hex editor). Maybe line-concatted tests can use platform-specific behaviour? Since I don't think this is a problem in practice.

raehik:
That looks to be it. I'll clarify the customised equality check and merge. A more complete solution would be to test purely where possible -- I had a look at this before, and it would entail a lot of work since the rewriter tests use file I/O as a basis. So I feel this is sensible for now.

PR #172, opened 2021-08-18
start block node refactor

raehik:


raehik:
See the relevant CamFort issue for rationale and updates. This works and could be used for most of the block constructs, allowing us to store more info and factor out a common pattern.

PR #173, opened 2021-08-20
store Declarators in CommonGroup

raehik:
* [ ] parser rule naming (bit all over the place)
  * [x] investigate/discuss impact for analysis, common block elimination

### DIFF
--- a/src/Language/Fortran/AST.hs
+++ b/src/Language/Fortran/AST.hs
@@ -528,6 +528,8 @@ data ImpElement a =
   | ImpRange        a SrcSpan String String
   deriving (Eq, Show, Data, Typeable, Generic, Functor)
 
+-- | Note that the 'Declarator's in common group definitions do not contain
+--   initializing expressions.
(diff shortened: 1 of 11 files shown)


raehik:
Somehow, I've broken an array initializer test for the F77 parser. The S/R conflicts went up from 1 to 4. Think it's due to declarator syntax allowing initializers. I specifically avoided that in some of the other parsers, but apparently works differently for F77 (and maybe F66).

raehik:
OK, so F66 doesn't have initializers in declarators, and the COMMON parsing in the other parsers uses explicit non-initializing declarators. This is prepared for merge, but I want to hold off until I've discussed the related CamFort change @ https://github.com/camfort/camfort/pull/136 with @dorchard .

dorchard:
This looks a good change.

raehik:
This is an AST change, but it has minimal impact on fortran-src & CamFort codebases. I'll clean up the new parser rules, add a changelog entry and merge.

PR #178, opened 2021-09-17
Properly include test data in package dist

raehik:
Need to sort this to pass Stackage's tests.


raehik:
(The Windows CI builds keep randomly failing due to some weird bug, I don't know what's going on. I *think* it keeps failing at building Alex, though.)

PR #180, opened 2021-09-24
Patch/fixup rewriter bug

### LINKED ISSUE #179: Rewriter Bug: '!' character in string literals treated as an explicit comment
While examining chunks, [`evaluateChunks_`](https://github.com/camfort/fortran-src/blob/37fe7b2005ff1bc1f4abec1020a783ce50b10666/src/Language/Fortran/Rewriter/Internal.hs#L194-L237) is too eager in determining where an explicit comment starts.

This is due to using `(BC.elemIndex '!' chStr)` will trip up on `!` characters inside string literals, and report that Sourcelocation as the start of an explicit comment.

The right way to go about this would be:
1. Traverse each chunk, while maintaining state about whether a character is in a string literal or not
2.  Use this state  to locate a `!` outside a string literal

The code in question lives here:

https://github.com/camfort/fortran-src/blob/37fe7b2005ff1bc1f4abec1020a783ce50b10666/src/Language/Fortran/Rewriter/Internal.hs#L209-L211

E.g.:

```
      write(8, *) "a really long string!", var1, var2, var3
                                       ^
                                       |__ `nextLen` points to here
```

envp:
Fixes #179 

[`evaluateChunks_`](https://github.com/camfort/fortran-src/blob/37fe7b2005ff1bc1f4abec1020a783ce50b10666/src/Language/Fortran/Rewriter/Internal.hs#L194-L237) is too eager in determining where an explicit comment starts.

This is due to using `(BC.elemIndex '!' chStr)` will trip up on `!` characters inside string literals, and report that Sourcelocation as the start of an explicit comment.

The right way to go about this would be:
1. Traverse each chunk, while maintaining state about whether a character is in a string literal or not
2.  Use this state  to locate a `!` outside a string literal

This PR fixes the above issue, and adds a failing test for detecting regressions.

### DIFF
--- a/src/Language/Fortran/Rewriter/Internal.hs
+++ b/src/Language/Fortran/Rewriter/Internal.hs
@@ -189,26 +189,29 @@ allChunks rchars = chunk : allChunks rest
 -- | Transform a list of 'Chunk's into a single string, applying
 -- continuation lines when neccessary.
 evaluateChunks :: [Chunk] -> ByteString
-evaluateChunks ls = evaluateChunks_ ls 0
+evaluateChunks ls = evaluateChunks_ ls 0 Nothing
 
-evaluateChunks_ :: [Chunk] -> Int64 -> ByteString
-evaluateChunks_ []       _       = ""
-evaluateChunks_ (x : xs) currLen =
-    if overLength
-    then "\n     +"
-         <> evaluateRChars xPadded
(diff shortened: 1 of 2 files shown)


raehik:
Nice catch, looks good. I *should* implement this in my continuation reformatter, but I think that has the opposite issue (it might split a comment without realising, and I don't believe comments can be continuated).

Thank you for the test case too! Merging.

PR #183, opened 2021-10-04
Constant folding for integer exponentation

raehik:


raehik:
Quick fix for an issue noting during discussion with @dorchard .

PR #184, opened 2021-10-14
Basic support for evaluating explicit constant expressions in type analysis

raehik:


raehik:
Old work replaced by #192 .

PR #188, opened 2021-10-28
Handle nonstandard kind parameter in type analysis

raehik:
Re note made by @poppysingleton earlier today.

There's some nonstandard syntax for defining a type's kind:

```fortran-free-form
INTEGER x*8
```

That's actually syntax that should only be used for `CHARACTER`s, and indicates their length. It should be invalid syntax for non-`CHARACTER` types. I'm unsure how nonstandard it is -- gfortran supports many nonstandard syntactic features, but I can't get it to accept this one.

Previously we prompted with a warning during type analysis. With this changeset, we assume it's the nonstandard kind syntax and prompt with an altered warning, then continue with it replacing any other kind parameter. Also adds a bunch of tests covering the main cases too.

### DIFF
--- a/fortran-src.cabal
+++ b/fortran-src.cabal
@@ -162,6 +162,7 @@ test-suite spec
       Language.Fortran.Parser.Fortran66Spec
       Language.Fortran.Parser.Fortran77.IncludeSpec
       Language.Fortran.Parser.Fortran77.ParserSpec
+      Language.Fortran.Parser.Fortran90PlusCommon
       Language.Fortran.Parser.Fortran90Spec
       Language.Fortran.Parser.Fortran95Spec
       Language.Fortran.Parser.UtilsSpec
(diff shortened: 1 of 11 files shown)


raehik:
fortran-vars includes the following test:

```fortran-free-form
INTEGER x*8(2)
```

This parses and gives the correct type. But fortran-src fails at parsing. It's down to the differences between the F77 parser (that fortran-vars uses) and the F90 (that I'm primarily testing on fortran-src). I'm about to add it to the rest of the parsers, but I'll try to confirm proper usage via specs first.

raehik:
Re above: Adding the nonstandard case to the F90 parser introduces 2 S/R conflicts, and I can see why (`EXPRESSION -> DATA_REF -> PART_REF` rule uses similar syntax). However, the tests pass. A safer fix could be to limit to integer literals in that case, which doesn't cause S/R conflicts.

raehik:
I think I'm actually for the integer lits-only change. Since this is nonstandard syntax anyway, and our F9X parsers are a lot more conforming than F66 and F77 (which had much less specified up front).

raehik:
I got it wrong, S/R conflicts are OK! Unlikely to cause parsing distress. As discussed with @dorchard , amending the rest of the parsers & merging

PR #189, opened 2021-11-02
CI: create & upload sdist, Haddocks

raehik:


raehik:
See https://github.com/camfort/camfort/pull/147 .

PR #193, opened 2021-11-17
parse REAL literals to convenient form

raehik:
We decompose REALs to their respective parts (significand & exponent),
but keep the numeric parts as strings to allow the user to select how to
represent the value. Along with a printer and parser, we also export a
"reader" to convert parsed REAL literals to any compatible Fortran type,
so for most uses users will only need that function to turn a Fortran
REAL literal into a Haskell value. (Previously, everyone had to
implement their own parsing.)


raehik:
This e.g. lets [Camfort](https://github.com/camfort/camfort/blob/2f5ac9a478116ae8c7aa2a5e079b81aa58988e45/src/Language/Fortran/Model/Translate.hs#L682) support the various Fortran-style real literals (instead of trying to read as a Haskell real), and [fortran-vars](https://github.com/camfort/fortran-vars/blob/e6db60bcd48ccf49f3de1cde584117f6a0c5f03f/src/Language/Fortran/Vars/Operation.hs#L49) not do any extra parsing work.

raehik:
@dominic would you be able to give this a quick look over? I'll bring it on Mon otherwise. It's largely adapted from some other code I wrote before but now gets called during parsing (actually, during lexing for free form, parsing for fixed form -- I stuck with the existing patterns)

raehik:
Oops, I'd already said I wasn't around Monday. I'm preparing other changes around this so it can all merge quick and smooth when we've discussed it.

raehik:
Discussed, giving it a merge.

PR #194, opened 2021-11-18
add early BOZ constant parsing, separate constructor

raehik:
In the same vein as #193 , we do what we can for BOZs as early as possible.

Thoughts and notes:

  * BOZs are now placed in their own constructor, but get parsed identically, which means they are alternatives in the `INTEGER_LITERAL` rule (in most of the parsers).
  * Could do a few more tests.
  * No special support for them in the rest of the codebase. If we're to follow the later Fortran standards, BOZ constants are untyped bitstrings, and their behaviour is pretty much all explicit rules like "works like this in this intrinsic function". Could check what fortran-vars deems useful if they have any relevant eval rules!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,21 +3,25 @@
     (thanks @envp) #179
   * CI builds on Mac; more release automation #181 #189
   * Handle nonstandard kind parameter in parsing & type analysis #188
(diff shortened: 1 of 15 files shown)


raehik:
Discussed. There's more code to consider adapting from fortran-vars for handling BOZs -- getting the AST change out of the way makes future changes easier, so merging.

PR #198, opened 2021-12-20
BOZ: +function to resolve as int

raehik:


raehik:
This targets most of [Vars.BozConstant](https://github.com/camfort/fortran-vars/blob/8e5e71268dcdeddb1f6ad992a1a8e885c95091b3/src/Language/Fortran/Vars/BozConstant.hs). The other functions in there are to do with evaluation and value type coercion (putting a Haskell `Integer` into a Fortran `INT(x)`), which are handled elsewhere in my ongoing type analysis.

PR #202, opened 2022-01-05
lex tabs as 4 spaces in fixed form

raehik:
Unsure about this. Might impact span positions in an unexpected way. Breaks a property test which needs rejigging with the new tab behaviour plugged (can't just look for 6 space prefix, we need to check for 1-tab-1, tab-2, 2-tab as well as 6)


raehik:
Dud change. I think I'll throw this away, it's almost entirely a UX thing which we can solve (less ideally) by instructing people to check for accidental tabs in their file. (This was triggered by learning that Notepad++ will insert tabs for Fortran indents, and makes it hard to change this setting.)

Issue #203, opened 2022-01-06
Copying prior to pass-by-reference triggered by parentheses, which we ignore.


dorchard:
This Tweet alerted me to something in Fortran that I was not aware of:
https://twitter.com/fortrantip/status/1479071485859962880

If an argument is passed to a subroutine/function that is a variable wrapped in parentheses, then a copy of the value is taken before applying pass-by-reference:

            module m
              implicit none
              contains
                subroutine double(i, i2)
                  integer :: i
                  integer :: i2
                  intent(in) :: i
                  intent(out) :: i2
                  print *, "incoming double, i = ", i
                  i2 = 2*i
                  print *, "outgoing double, i = ", i
                end subroutine double
            end module m
            
            program copy
              use m
              implicit none
              integer :: j, z
              j=3
              call double(j, j)
              print *, "(1b) in main, j = ", j
            
              call double((j), j)
              print *, "(2) in main, j = ", j
            
            end program copy

We have a problem with this as our parser strips away extra parentheses:
https://github.com/camfort/fortran-src/blob/master/src/Language/Fortran/Parser/Fortran90.y#L1027
However, this appears to be valid Fortran 90 code. I haven't been able to pin down when this was introduced into the language, but we should certainly do something about this!

raehik:
Interesting. I certainly have some misconceptions about binders to variables, I'm fuzzy on rules around temporaries and dummys (are they the same?).

I don't like it, but I can't think of anything other than an `ExpBracket Expression` constructor for parsing. A post-parse pass would remove them, and annotate some constructors (or change them to a different one) where required. More concretely, we could do:

```haskell
{-# LANGUAGE DataKinds #-}
data Expression = ExpValue (p :: Polarity) | ExpBracket Expression | ... -- only ExpValue has polarity
data Polarity = DirectValue | IndirectExpression

-- parser always parses @ExpValue 'DirectValue@, and parses brackets explicitly as @ExpBracket e@

collapseBrackets :: AST -> AST
-- * rewrite @ExpBracket (e :: ExpValue 'DirectValue)@ to @ExpValue 'IndirectExpression@
-- * remove all other @ExpBracket@s -- so none in AST after pass
```

[...]

PR #204, opened 2022-01-11
Parse bracketed plain vars in a function call specially e.g. `call((a), a)`

raehik:
Targets #203 .

Based on the following tweet and referenced material

    call double((i), i)

has potentially different meaning to, and should be parsed differently
from

    call double(i, i)

As I understand, variables may be used as expressions in two ways:
either directly, as a value access, or indirectly, as an expression to
be evaluated. I add a new Expression constructor for the latter, throw
some hopefully-overriding parser rules in and hope for the best.

https://twitter.com/fortrantip/status/1479071485859962880


raehik:
@dorchard Seems I get away with S/R conflicts by simply adding an "override" case to `EXPRESSION` parser rules, without refactoring. This change would mean adding another case for all `Expression` pattern matches, which should almost always be treated the same way as an `ExpValue (ValVariable name)`.

I wonder if I should instead move this to `Argument`. Make an `Either ExpValSpecial Expression` type and use it in `Argument`. Then the changes are limited to the `call` context only, and don't muddy the AST. Thinking about it I've now become fairly certain that's better, I didn't catch that `Argument` existed. Will try.

raehik:
It looks like we get away with it in all parsers. I need to add some explicit tests, and think about how we handle the new case of "variable as expression" in analyses. Should be in the dataflow analysis.

raehik:
Note: This introduces 2 shift/reduce conflicts in each parser (expected).

raehik:
Merging the parser change & addressing behaviour (special behaviour with aliasing in certain functions) in another PR.

PR #205, opened 2022-01-31
restructure parsers, various refactoring

raehik:
I needed to shift some parser code around for benchmarking, so I could configure everything nicely. I took the chance to do a small rewrite I'd wanted to do for a while. The main changes:

  * User-facing parsers are exposed in a new `Language.Fortran.Parser` module. This means a single place for various shared parsing decisions/documentation: transformations, initialization, common configurations for parsers.
    * This also unwinds some awkward dependency cycles. Parsers now have minimal dependencies, and won't constantly rebuild with changes to unrelated code.
  * Happy parser modules now expose only their parsers. All shared code is moved to various common modules.
  * `data Transformation` has been removed. Transformations are now plain functions with a concrete type and passed around as such.
  * Code deduplication, documenting unusual behaviours.

I also made a bunch of language extensions default. Nothing that can change behaviour. My criteria was one of:

  * constantly used (BangPatterns, TupleSections)
  * deriving-related
  * should be default anyway (FlexibleX, MPTC)

Things that may warrant discussion:

  * [ ] I renamed various modules for clarity. Things like `ParserMonad` to `Parser.Monad`, grouping parsers with their lexer under `Parser.Fixed.*`, `Parser.Free.*`. Is this OK? (I figured there are enough breaking changes to go all in.)
  * [ ] Speaking of breaking changes, we could make a bunch of shim modules to retain some backwards compatibility. I'm not certain of the value, though. `ParserMonad` had long kept too much unrelated code together, so this becomes a useful opportunity to require changing it.


raehik:
Merging. Briefly gone over with @dorchard and we can tinker further before a release as needed.

PR #206, opened 2022-02-01
Handle call-by-value parameters

### LINKED ISSUE #203: Copying prior to pass-by-reference triggered by parentheses, which we ignore.

raehik:
`Analysis.BBlock.perBlock` is where the decision is made on how to create blocks from `call()`s. Previously, it would check to see if the expression was assignable: either a plain variable, or a subscript. If so, it would "re-assign" them (I'm not completely clear on this part). Now, we treat bracketed plain variables as non-assignable, figuring that they're used as call-by-value.

Closes #203 .

TODO

  * [ ] `perBlock`: do same for the expression function call (pull the pattern out)
  * [ ] write some tests
  * [ ] document a bit clearer?

### DIFF
--- a/src/Language/Fortran/AST.hs
+++ b/src/Language/Fortran/AST.hs
@@ -485,14 +485,23 @@ data Use a =
   | UseID a SrcSpan (Expression a)
   deriving (Eq, Show, Data, Typeable, Generic, Functor)
 
--- TODO potentially should throw Maybe String into ArgumentExpression too?
-data Argument a = Argument a SrcSpan (Maybe String) (ArgumentExpression a)
+data Argument a = Argument a SrcSpan
+                           (Maybe Name)             -- ^ optional @var = ...@
+                           (ArgumentExpression a)   -- ^ expression (wrapped)
   deriving (Eq, Show, Data, Typeable, Generic, Functor)
 
 data ArgumentExpression a
   = ArgExpr              (Expression a)
   | ArgExprVar a SrcSpan Name
   deriving (Eq, Show, Data, Typeable, Generic, Functor)
+
+instance Spanned (ArgumentExpression a) where
+    getSpan = \case
+      ArgExpr            e -> getSpan e
+      ArgExprVar _a  ss _v -> ss
+    setSpan ss = \case
+      ArgExpr            e -> ArgExpr $ setSpan ss e
+      ArgExprVar  a _ss  v -> ArgExprVar a ss v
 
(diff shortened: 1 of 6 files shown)


raehik:
I'm testing like so:

```fortran
module m
    implicit none
    contains

    subroutine double(i, i2)
        integer, intent(in)  :: i
        integer, intent(out) :: i2
        print *, "entering double, i =", i
        i2 = 2*i
        print *, "leaving  double, i =", i
    end subroutine double

end module m

program overlapping_arg
    use m, only: double
    implicit none
    integer :: j
    j = 3
    call double(j, j)
    print *, "(1) in main, j =", j
    call double(j, j)
    print *, "(2) in main, j =", j
end program overlapping_arg
```

And running `stack run -- test.f90 --show-flows-from=B13 --dot | dot -Tpng > graph.png`. Bracketing various `j`s in the 2 calls to `double` turns that usage to a call-by-value, which effectively removes that line on the data flow graph.

raehik:
I ignored the named parameter feature that Fortran has! Almost finished re-jigging the parsers to properly handle that. (I don't think we look at named parameters in the analysis, so I'm not fixing anything there.)

raehik:
Pending while we resolve some potential issues with the `--show-flows-{to,from}` flags -- they give unexpected output, which could be flipped directions or just ambiguous naming

PR #209, opened 2022-02-22
better restrict COMPLEX literal parsing

raehik:
COMPLEX literals were previously stored as `ValComplex (Expression a) (Expression a)` and parsed similarly, like an expression tuple. The F90 and F77 standards both state the constituent parts are literals, with F77 allowing reals only and F90 allowing both integer and real literals (and apparently performing delayed casting). This change reflects the F90 form in the constructor in every parser. Should make it easier to use COMPLEX values.


raehik:
Checked parsing some Fortran projects I've got on my drive, everything works as expected. Merging!

PR #210, opened 2022-03-02
Fix include parsing

RaoulHC:
Block grammar rules expect to end with a newline, which isn't
necessarily the case for an include, so we should append a newline
to ensure it's correctly parsed.


raehik:
Thank you! Looks like I put the newline in the wrong place when moving this over.

raehik (approved):

PR #213, opened 2022-03-08
export BOZ as sized two's complement integer

raehik:
Fortran integers are stored in two's complement (in my understanding),
so this is a useful definition for converting BOZ strings to numeric
types.

Also clarified Boz equality by shoving the "check equality while
ignoring conformance flags" into the Eq instance. It's for testing
syntactic equality and shouldn't get used much in the first place, so
fewer definitions seems sensible.


raehik:
brought up by @RaoulHC some months ago, and I very eventually recognized the nice general solution. `int(z'FFFFFFFF', 4) == -1`. with easier numbers, `int(z'FF', 1) == -1`, `int(z'FF', 2) == 255`

PR #216, opened 2022-04-24
Add support for array specs in fortran 77 cray pointers

ProgramFan:
This PR adds a parsing rule to handle array specs in cray pointers in fixed-form fortran. The spec is from https://docs.oracle.com/cd/E19957-01/805-4941/z40000a54ba7/index.html. Basic tests with a simple fortran 77 source with `-v f77l` passed.

### DIFF
--- a/src/Language/Fortran/Parser/Fixed/Fortran77.y
+++ b/src/Language/Fortran/Parser/Fixed/Fortran77.y
@@ -607,7 +607,9 @@ POINTER_LIST :: { [ Declarator A0 ] }
 | POINTER                  { [ $1 ] }
 
 POINTER :: { Declarator A0 }
-: '(' VARIABLE ',' VARIABLE ')'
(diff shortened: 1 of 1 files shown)


dorchard:
Thanks for this! Would it be possible for you to put a comment in here to explain that this is for Cray pointers and also to add an example as a test in here: https://github.com/camfort/fortran-src/blob/master/test/Language/Fortran/Parser/Fixed/Fortran77/ParserSpec.hs?

ProgramFan:
Thanks for the review. I am preparing it.

ProgramFan:
I agree with the idea of making separate constructors for cray pointers, since they can appear in both fixed and free form sources. It would also solve the awkward situation here that I need to workaround the limitation of `Declarator`. Unfortunately, I am still new to haskell, so I am afraid I am not capable of making such complex changes.

PR #218, opened 2022-05-06
use generic-optics to define First,SecondParameter

raehik:
Our generics code was unsafe in that if you tried to derive an instance
which was invalid (e.g. you say an `a` is the first parameter in each
constructor, but it isn't), it would still derive an instance, but give
you an `undefined` runtime error upon using it. Pushing that error to
compile time is complex, but it's a solved problem by kcsongor's in his
wonderful generic-lens library.

This change adds generic-optics and optics-core to dependencies. They're
very lightweight. Performance should be the same; compile time should
most likely increase.

### DIFF
--- a/fortran-src.cabal
+++ b/fortran-src.cabal
@@ -144,7 +144,9 @@ library
     , either >=5.0.1.1 && <5.1
     , fgl ==5.*
     , filepath ==1.4.*
+    , generic-optics >=2.2.1.0 && <2.3
     , mtl >=2.2 && <3
+    , optics-core >=0.4.1 && <0.5
     , pretty >=1.1 && <2
     , temporary >=1.2 && <1.4
     , text >=1.2 && <2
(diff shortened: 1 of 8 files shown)


raehik:
Looks like the compile time increase is *much* bigger than I thought, and memory too. The main module it should affect is `Language.Fortran.AST`. Running `stack ghci` on master and this branch and timing that module (mem% is from htop) shows:

  * master: 17.0% mem, 10s
  * this branch: 44.7% mem, 40s

Bit unsure now. I wonder if it would improve if `Expression` weren't so mutually recursive.

raehik:
The extra work generic-lens has to do for safe derivations is just too much. At least it's shown me another questionable instance to remove. I'm replacing this PR with one which doesn't alter the generic derivation over here: https://github.com/camfort/fortran-src/pull/219

PR #220, opened 2022-05-09
add a handful of record selectors

raehik:


raehik:
Simple refactor to improve quality of generated Aeson instances.

PR #221, opened 2022-05-10
refactor DoSpec to not use Statement

raehik:
This removes `Expression`'s mutual recursion with `Statement`.

### DIFF
--- a/src/Language/Fortran/AST.hs
+++ b/src/Language/Fortran/AST.hs
@@ -588,7 +588,7 @@ data FlushSpec a =
   deriving (Eq, Show, Data, Typeable, Generic, Functor)
 
 data DoSpecification a =
-  DoSpecification a SrcSpan (Statement a) (Expression a) (Maybe (Expression a))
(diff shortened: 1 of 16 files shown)


raehik:
I'm getting test failures related to the block analysis. uniplate behaviour has changed because with `DoSpec`s no longer storing a `Statement`, it doesn't get picked up by `universeBi`. I'm trying to convert `DoSpec`s to `StExpressionAssign`s to remove the need to rewrite the traversals or reconsider the AST, but it's not working.

raehik:
Another solution would be to change `ExpImpliedDo` instead. But `DoSpecification` shouldn't store a `Statement`, the only thing allowed in there is `StExpressionAssign`.

raehik:
The 5 failing tests are all related to `DataFlow.reachingDefinitions`, which calls `allLhsVars`, which were the problematic functions. Will spend a bit more time on this today... but if I can't solve it, I'll leave it for now.

raehik:
bit lost, only 2 changes seem to have any impact on test results. Leaving for now.

PR #224, opened 2022-05-13
remove DoSpec from ExpImpliedDo

raehik:


raehik:
gfortran has some strange behaviour for implied DO loops:

  * used in a `print`, the final value of the loop variable is retained
  * used in an assignment, the loop variable is left untouched

```fortran
program main
    integer i
    integer is(5)
    print *, i                  ! 0
    print *, ( i, i = -1, 2 )
    print *, i                  ! 3
    is = [ (i, i = 2, 6) ]
    print *, is
    print *, i                  ! 3
end
```

Putting aside the `print` behaviour, they feel like syntactic sugar. gfortran even tells you if it's the wrong shape for the assigning array, implying it evaluates them to arrays during compilation.

dorchard:
The tricky thing is that now we've lost the data flow path due to an assignment...!

mrd:
I wouldn't count on gfortran for sensible/reliable behaviour.

PR #225, opened 2022-06-10
Set the span correctly for assumed size arrays

RaoulHC:
These only used the span of the star and not the surrounding brackets
which meant the the span of the declarations missed the closing bracket.
Also noticed we missed the correct span for one of the implied do cases.
Looks like the other standards parsers didn't miss this.

We also don't have any tests for checking `SrcSpan`s, not sure how worth it is to add them.


raehik (approved):
Thanks. Inspecting this, it's the situation same as the bracketed expressions rule in `EXPRESSION`, where we have to expand the source span to the brackets.
I wouldn't worry about source span tests for now. I think the long-term approach would be to slowly refactor parser tests to include source span checks too. No doubt it's been avoided because it's just awkward.

PR #227, opened 2022-06-15
Add entry point for parsing includes

RaoulHC:
We're refactoring an issue that occurs in include files, and exposing this allows us to directly refactor them.

We may need in future to try and perform some amount of analysis on them (which should doable I believe), but this is sufficient for now.

### DIFF
--- a/src/Language/Fortran/Parser.hs
+++ b/src/Language/Fortran/Parser.hs
@@ -36,6 +36,7 @@ module Language.Fortran.Parser
   -- * F77 with inlined includes
   -- $f77includes
   , f77lIncludes
+  , f77lIncIncludes
   ) where
 
 import Language.Fortran.AST
@@ -252,6 +253,18 @@ f77lIncludes incs mods fn bs = do
                                  (defaultTransformation Fortran77Legacy)
                                  pf''
         return pf'''
+
+-- | Entry point for include files
+-- 
+-- We can't perform full analysis (though it might be possible to do in future)
+-- but the AST is enough for certain types of analysis/refactoring
+f77lIncIncludes
+  :: [FilePath] -> String -> B.ByteString -> IO [Block A0]
+f77lIncIncludes incs fn bs =
+  case makeParserFixed F77.includesParser Fortran77Legacy fn bs of
+    Left e -> liftIO $ throwIO e
+    Right bls ->
+      evalStateT (descendBiM (f77lIncludesInline incs []) bls) Map.empty
 
 f77lIncludesInner :: Parser [Block A0]
 f77lIncludesInner = makeParserFixed F77.includesParser Fortran77Legacy


src/Language/Fortran/Parser.hs
```diff
@@ -253,6 +254,18 @@ f77lIncludes incs mods fn bs = do
...
+-- | Entry point for include files
+-- 
+-- We can't perform full analysis (though it might be possible to do in future)
+-- but the AST is enough for certain types of analysis/refactoring
+f77lIncIncludes
+  :: [FilePath] -> String -> B.ByteString -> IO [Block A0]
+f77lIncIncludes incs fn bs =
+  case makeParserFixed F77.includesParser Fortran77Legacy fn bs of
+    Left e -> liftIO $ throwIO e
+    Right bls ->
```
raehik, line 266: `Parser`s spit out a `ProgramFile`, so this is misleading. The function can return a `[Block A0]` at the end because uniplate applies its biplate magic.

raehik (commented):
If what you want is a no-transform include parser, I'm inclined to refactor a bit:

```haskell
f77lIncludes'
    :: [FilePath] -> String -> B.ByteString
    -> IO (ProgramFile A0)
f77lIncludes incs fn bs = do
    case f77lNoTransform fn bs of
      Left e -> liftIO $ throwIO e
      Right pf -> do
        let pf' = pfSetFilename fn pf
        evalStateT (descendBiM (f77lIncludesInline incs []) pf') Map.empty

f77lIncludes
    :: [FilePath] -> ModFiles -> String -> B.ByteString
    -> IO (ProgramFile A0)
f77lIncludes incs mods fn bs = transform <$> f77lIncludes' incs fn bs
  where
    transform = runTransform (combinedTypeEnv mods)
                             (combinedModuleMap mods)
                             (defaultTransformation Fortran77Legacy)
```

RaoulHC:
> is this to allow parsing with includes without doing post-parse transformations? Do the transformations make things more awkward?

It's to refactor include files directly. There's no issue with doing transformations, though provided the includes are only non-executable statements (they don't *have* to be, but they would be a massive pain to use if they include any executable statements), there is no need for transformations.

src/Language/Fortran/Parser.hs
RaoulHC: I don't think that's true? `makeParseFixed` returns whatever it's first argument returns, which in the case of `includesParser` is `[Block A0]`.

RaoulHC: At the very least I've tried running this entry point on include files and it works, which it wouldn't if it were expecting to parse `ProgramUnit`s

raehik:
Thanks, I get it. It looked similar to other existing code, but now I see includes work with `[Block]`s and use different parsers, so it's sensible.

PR #228, opened 2022-06-21
Fix block parsing

RaoulHC:
The way it currently worked meant it would fail to parse includes that
didn't end with a newline, which is valid. To allow this the block
parsing had to be reorganized, with the block parser not expecting
trailing newlines, but otherwise behaves the same.

We might want to do this for other parsers, but because they currently
don't expand includes, it's not actually a problem for them.

### DIFF
--- a/src/Language/Fortran/Parser/Fixed/Fortran77.y
+++ b/src/Language/Fortran/Parser/Fixed/Fortran77.y
@@ -194,16 +194,16 @@ PROGRAM_UNITS :: { [ ProgramUnit A0 ] }
 | maybe(LABEL_IN_6COLUMN) PROGRAM_UNIT maybe(NEWLINE) { [ $2 ] }
 
 PROGRAM_UNIT :: { ProgramUnit A0 }
-: program NAME NEWLINE BLOCKS ENDPROG
-  { PUMain () (getTransSpan $1 $5) (Just $2) (reverse $4) Nothing }
-| TYPE_SPEC function NAME MAYBE_ARGUMENTS NEWLINE BLOCKS ENDFUN
-  { PUFunction () (getTransSpan $1 $7) (Just $1) emptyPrefixSuffix $3 $4 Nothing (reverse $6) Nothing }
-| function NAME MAYBE_ARGUMENTS NEWLINE BLOCKS ENDFUN
-  { PUFunction () (getTransSpan $1 $6) Nothing emptyPrefixSuffix $2 $3 Nothing (reverse $5) Nothing }
-| subroutine NAME MAYBE_ARGUMENTS NEWLINE BLOCKS ENDSUB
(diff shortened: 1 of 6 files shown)


src/Language/Fortran/Parser/Fixed/Fortran77.y
```diff
@@ -236,38 +236,36 @@ MAYBE_ID :: { Maybe Name }
...
-| STATEMENT NEWLINE { BlStatement () (getSpan $1) Nothing $1 }
-| COMMENT_BLOCK { $1 }
+: IF_BLOCK { $1 }
+| LABEL_IN_6COLUMN STATEMENT { BlStatement () (getTransSpan $1 $2) (Just $1) $2 }
+| STATEMENT { BlStatement () (getSpan $1) Nothing $1 }
+| comment { let (TComment s c) = $1 in BlComment () s (Comment c) }
 
 IF_BLOCK :: { Block A0 }
-:                  if '(' EXPRESSION ')' then NEWLINE BLOCKS ELSE_BLOCKS
+: if '(' EXPRESSION ')' then BLOCKS NEWLINE ELSE_BLOCKS
```
raehik, line 253: I think I'm misunderstanding something, but when I read this (and the `BLOCKS`, `BLOCK` rules), I don't see where the initial `NEWLINE` between `then` and the internal blocks gets consumed (which it must do, because all the tests are happy!).

raehik (approved):
`Language.Fortran.Parser.f77lIncludesInline` snocs a newline onto include files before parsing them with a note about this behaviour you've fixed. We should update the comment there (and remove the snoc as long as it has no unexpected consequences).

raehik:
Cheers for this. I had felt the previous design was a bit of a code smell, swapping `unlines` for `intercalate "\n"` feels like an improvement.
Have made a note for pulling this into the other parsers #229 .

src/Language/Fortran/Parser/Fixed/Fortran77.y
RaoulHC: `BLOCKS` can be empty, so `BLOCKS NEWLINE BLOCK` can be `{- empty -}, newline, block`, hence everything being happy.

raehik: Oh! It's caught by `BLOCKS NEWLINE BLOCK`, where the first `BLOCKS` is the empty case. No S/R or R/R conflicts, so seems good. I wonder what this might mean for performance though.

test-data/f77-include/foo.f
```diff
@@ -1 +1 @@
-      integer a
+      integer a
```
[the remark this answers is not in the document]
RaoulHC: Added an extra test so it does both these cases.

PR #232, opened 2022-07-21
Rewriter Fixes

jasonqxu:
Wrapping implicit comments - contributed by Anthony Burzillo
* Add test with incorrect wrapping for comments
* Add test for continuation comments
* Add failing test related to the context of coments after replacement
* Fix comments after forced continuations
* Refactor chunk expansion with documentation
* Add test for inline comments going over 72

Fix incorrect line wrap from eager comment detection - contributed by Vaibhav Yenamandra
* Track quotations between chunks while searching
* When looking for a '!' in source, search outside strings

### DIFF
--- a/src/Language/Fortran/Rewriter.hs
+++ b/src/Language/Fortran/Rewriter.hs
@@ -29,17 +29,18 @@ where
 import qualified Data.ByteString.Lazy.Char8    as BC
 import qualified Language.Fortran.Rewriter.Internal
                                                as RI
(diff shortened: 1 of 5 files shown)


jasonqxu:
The two fixes in this PR were implemented after the integration of fixed-form rewriter into `fortran-src`.

raehik:
Nice fixes, thank you Jason for the upstream! Merging (Windows CI is failing with an unrelated issue).

PR #233, opened 2022-08-08
Make sure non literals have parens where needed

RaoulHC:
Otherwise pretty printing declarations results in invalid fortran such as `foo**` rather than `foo*(*)`.


raehik:
Thanks Raoul, nice catch. I noted `noParensLit` was used in multiple places, so I pulled it out to top level (and did some other half-related tweaks I remembered). I'm happy to merge if you are.

RaoulHC:
LGTM, can be merged once CI's passed.

PR #234, opened 2022-08-17
Make sure inlined includes aren't reversed

RaoulHC:
Also added some tests so this is caught in future.

Not sure how I didn't notice this before, but I guess the tool I was using this ( change for wasn't generating a symbol table, which falls over when includes are in reverse order.

### DIFF
--- a/src/Language/Fortran/Parser/Fixed/Fortran77.y
+++ b/src/Language/Fortran/Parser/Fixed/Fortran77.y
@@ -237,7 +237,7 @@ MAYBE_ID :: { Maybe Name }
 NAME :: { Name } : id { let (TId _ name) = $1 in name }
 
 INCLUDES :: { [ Block A0 ] }
-: BLOCKS maybe(NEWLINE) { $1 }
+: BLOCKS maybe(NEWLINE) { reverse $1 }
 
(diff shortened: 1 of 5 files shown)


RaoulHC:
Huh, not sure why the Windows CI would fail, any ideas @raehik?

raehik:
Nice catch. In my experience, Windows CI fails are usually to do with newline behaviour. But on the CI logs, the failing tests seem to have matching `Show` outputs. My bet is the `Position`s in the `SrcSpan`s have different absolute offsets. Where the actual issue is in all this, not sure.

All good to merge anyway. Happy to drop a CPP clause on the failers for Windows. I'm taking a look now, but imagining I can't do much. (Wonder if `unlines` lies on its documentation, and actually inserts `\r\n` on Windows?)

raehik:
My local Windows setup *does* pass all the tests. I guess the CI setup doesn't use MSYS2, so likely behaviour differences... but I would file this under the "problems due to `Eq` inspecting `SrcSpans`" folder and forget about it! I'll push a fix to disable the problematic tests on Windows with a note.

RaoulHC:
Yeah I did wonder if it was an `\r\n` issue, but wasn't sure where it might be coming from.

PR #237, opened 2022-08-26
Catch more legacy comments

### LINKED ISSUE #236: Legacy lexer doesn't catch some comments
```fortran
      program main
      ! I'm a comment
c     I'm also a comment
      end program main
```
In the above example only the second comment is caught as a block comment and the first is ignored. Afaict this should parse fine (see [here](https://github.com/camfort/fortran-src/blob/master/src/Language/Fortran/Parser/Fixed/Fortran77.y#L251)) but presumably the fixed lexer ends up ignoring this form of comment. This type of comment is parsed as a block comment in the fortran 90 lexer/parser.

Tested on the latest master b6a7f8e.

RaoulHC:
Three changes to the fixed lexer:
* Catch inline comments when they're the only thing on a line.
* Catch them case sensitively. Not sure why this wasn't the case before but information in the comments might be lost otherwise.
* Catch parts of comments that go over column 72. Once again there might be information that's lost otherwise.

Resolves #236.

### DIFF
--- a/src/Language/Fortran/Parser/Fixed/Lexer.x
+++ b/src/Language/Fortran/Parser/Fixed/Lexer.x
@@ -530,6 +530,12 @@ setCaseInsensitive = do
   ai <- getAlex
   putAlex $ ai { aiCaseSensitive = False }
 
+setInComment :: LexAction ()
+setInComment = getAlex >>= \ai -> putAlex ai { aiInComment = True }
+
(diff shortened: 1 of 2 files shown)


raehik (approved):
Thanks for fixing this! The case sensitive comments and parsing beyond column 72 are nice too.

PR #242, opened 2022-09-16
Fix pad implicit comments

RaoulHC:
It was getting run for RChar removals, which are run in per character chunks and ended up adding a line with a single character as an implicit comment with no newline and breaking the source.

### DIFF
--- a/src/Language/Fortran/Rewriter/Internal.hs
+++ b/src/Language/Fortran/Rewriter/Internal.hs
@@ -267,20 +267,22 @@ evaluateChunks_ (x : xs) currLen quotation =
   -- Text after line 72 is an implicit comment, so should stay there regardless
   -- of what happens to the rest of the source
   padImplicitComments :: Chunk -> Int -> Chunk
-  padImplicitComments chunk targetCol =
-    let zippedChunk = zip [0 ..] chunk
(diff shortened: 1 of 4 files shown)


raehik (approved):
Thanks for this! One note: `isMarkedForRemoval` threw me for a bit due to the chunking. Can a removal chunk ever look different from `[RChar _ True _ _]`? The `Chunk` docs imply not:

```haskell
-- | As we advance through the ['RChar'] list, we consider "chunks"
-- as the unit of text written out. A chunk is either:
--
--     1. original source text up to a newline character, end of file
--        or 'RChar' described in 2.
--     2. a single 'RChar' that has non-empty replacement string
--        or is deleted.
type Chunk = [RChar]
```

I'm happy with the implementation being safer, but that function might belong as a top-level def.

RaoulHC:
Forgot to reply to this, but yeah I *think* it can only be that but wasn't sure so just looked for any `RChar` marked as replaceable. I'm not sure what would happen if it did encounter a mixed chunk. There probably could be a better representation of  the replacement chunks tbh, to avoid this kind of ambiguity but I was just looking to fix the existing issue.

I'll move `isMarkedForRemoval` to the top level though.

PR #244, opened 2022-10-12
SemType: replace Dimension to support assumed-size

raehik:
The `TArray` constructor in `SemType` supports "fully-dynamic" arrays and "fully-static" ones, but not assumed-size arrays, which are a holdover from older Fortran standards where only the final dimension is dynamic. We replace the `Dimension` type synonym with a list-like data type in order to support them.

### DIFF
--- a/src/Language/Fortran/Analysis/SemanticTypes.hs
+++ b/src/Language/Fortran/Analysis/SemanticTypes.hs
@@ -2,7 +2,7 @@
 
 module Language.Fortran.Analysis.SemanticTypes where
 
-import           Data.Data                      ( Data, Typeable )
+import           Data.Data                      ( Data )
 import           Control.DeepSeq                ( NFData )
 import           GHC.Generics                   ( Generic )
 import           Language.Fortran.AST           ( BaseType(..)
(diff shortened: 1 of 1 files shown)


raehik:
@RaoulHC the type change should look something like this -- push and make whatever changes you want here (or make another PR), though most work will be in fortran-vars and other pkgs

src/Language/Fortran/Analysis/SemanticTypes.hs
```diff
@@ -35,10 +35,8 @@ data SemType
   | TCharacter CharacterLen Kind
   | TArray SemType (Maybe Dimensions) -- ^ Nothing denotes dynamic dimensions
```
RaoulHC, line 36: Should this be changed to `TArray SemType Dimensions`? A single dynamic dimension can just be represented by `DimensionsFinalStar` now.

raehik: Good point. I had thought the `Nothing` option enables declaring an assumed-shape array type (e.g. `integer arr(:)`), but I think I misunderstood them, since the Fortran 2018 spec implies that they tell you the rank/no. of dimensions.

We could change to `TArray SemType (Either Int Dimensions)` to encode assumed-shape arrays. But I'm leaning towards simplifying as you recommend and leaving a comment on assumed-shape arrays. They don't show up until later specs.

RaoulHC: Ah right, yeah I can't say I'm familiar with any newer way of declaraing assumed size arrays that those standards might allow. If it's only from Fortran 2018 then given fortran-src doesn't support that at all it probably makes sense to leave it for now.

raehik: It's not from Fortran 2018, that was just the standard I had on hand. Apparently in F90. I'm still shying away from explicitly encoding them for now, since `SemType` tends to deal with F77 code (expanding that leads to extra complexity).

RaoulHC (approved):

raehik:
Added a `Dimensions -> Maybe [(Int, Int)]` function for compatibility, and fixed laziness (matches lists now).

src/Language/Fortran/Analysis/SemanticTypes.hs
RaoulHC: Oh hmm, I guess at some point we want to move to something a bit more f90 compliant but that's probably still a while off and can be expanded then (or perhaps sooner if anyone else wants f90 analysis from this).

PR #245, opened 2022-10-17
Don't inline solo includes

RaoulHC:
When dealing with solo includes we don't have the context to know where its includes are so we probably shouldn't be inlining them.

Happy for this to be named something else. 
I'm also unsure if there are cases where we want to inline includes still, but I don't think we'll have the context for our use cases, but if you can think of one maybe it will make sense for us to have a function that inlines them and one that doesn't.


RaoulHC:
(Will update [this](https://github.com/camfort/fortran-src-extras/blob/main/src/Language/Fortran/Extras.hs#L94-L97) in fortran-src-extras once this has gone through)

RaoulHC:
Yeah that would work, probably makes sense to put that definition in fortran-src-extras then.

raehik:
Nice, thanks for this. I'll merge this and bump fortran-src-extras.

PR #248, opened 2022-11-03
Line directive pragma handling in freeform lexer

mrd:
For the freeform lexer, fixes longstanding problems with lexing line directives (both inside and outside of continuations). Handles directives of the following forms:

- `# 40 "file.f"`
- `# line 40 "file.f"`

### DIFF
--- a/src/Language/Fortran/Parser/Free/Lexer.x
+++ b/src/Language/Fortran/Parser/Free/Lexer.x
@@ -1050,6 +1050,8 @@ skipContinuation ai' = _skipCont ai' (0::Integer)
         then _advance ai 3
         else if _curChar == '!'
         then _advance ai 2
+        else if _curChar == '#'
+        then _pragma ai ""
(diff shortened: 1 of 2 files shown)


src/Language/Fortran/Parser/Free/Lexer.x
```diff
@@ -1093,22 +1103,29 @@ advance move position =
...
 
--- Handle pragmas that begin with #
-lexHash :: LexAction (Maybe Token)
-lexHash = do
-  ai <- getAlex
-  m <- getMatch
-  case words (drop 1 m) of
+processLinePragma :: String -> AlexInput -> AlexInput
+processLinePragma m ai =
+  case dropWhile ((`elem` ["#", "line", "#line"]) . map toLower) (words m) of
```
RaoulHC, line 1108: Won't this happily process lines like `# line line #line # line 40 "file.f"`? Which it probably should complain about.

PR #251, opened 2022-11-18
Fix reformatting of 73 character long lines

ksromanov:
According to http://fortranwiki.org/fortran/show/Continuation+lines ampersand sign should be placed at column 73 and after, but not at column 72. So that `&` is ignored in fixed-form and is processed in free-form languages.

This patch fixes old behavior, which sometimes produced lines 73 characters long.

We (D. Beer, R. Hidalgo-Charman and myself) found this bug working on random FORTRAN code autogeneration tool. One of the examples of generated files with too long lines:

```fortran
.........
      if (- (+ (+ (50 / 38)))) then
        x = + (((y - (z - 48)) * ((z + 62) - z)) + (((- z) * x) / (- x)))
        x = 53
        print *, 24
        print *, + (z - (+ (+ (+ 20))))
        x = + y
        x = + (+ x)
        x = + (42 - (((- 82) / (y - y)) / 45))
        print *, (((- (z / 86)) * ((y + y) - (44 + x))) / (- ((- y) - (-&
     & x)))) - z
        x = + z
      else if (z) then
.........
```

### DIFF
--- a/src/Language/Fortran/PrettyPrint.hs
+++ b/src/Language/Fortran/PrettyPrint.hs
@@ -1134,7 +1134,11 @@ data ReformatState
 -- Ensures that no non-comment line exceeds 72 columns.
 --
 -- The reformatting should be compatible with fixed and free-form Fortran
--- standards. See: http://fortranwiki.org/fortran/show/Continuation+lines
+-- standards, so called `intersection` format. In this format the
+-- last statement character should not be placed after column 72, however
+-- continuation character should be put at column 73 (it should be ignored in
+-- fixed-form language)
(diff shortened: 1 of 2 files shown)


RaoulHC (approved):
LGTM, but I'll let @raehik have a look as well.

raehik:
Sorry for the delay in looking at this. Yep, the lookahead I did meant that exactly 73-character lines wouldn't get split. Over-eager splitting doesn't happen here due to newline handling happening earlier. Thanks for the fix and links to docs! Happy to merge.

raehik (approved):

PR #253, opened 2023-01-12
replace Fortran type/value representation with non-GADTs; replace constant folder

raehik:
The data types for representing Fortran types in `Language.Fortran.Repr` used GADTs and type-level trickery to enable reasoning about different (Fortran) kinds of the same Fortran types soundly, without too much extra work. This ended up not meshing with other parts of fortran-src, which does a lot via `Data`/Scrap Your Boilerplate, which is a real pain with GADTs.

Eventually, it was decided to scrap the extra type safety. This PR does that, and means we can replace the INTEGER-only constant folder with a call out to the expression evaluator.

The original PR description follows:

---

This is how we should resolve the `Binary`/`Out` instances issue, preventing placing `Repr` types into the `Analysis a` wrapper.

  * GADTs need special support
    * `Out` via `Show` (`DerivingVia`)
    * `Binary` -- unsure right now. probably just manually expand the generics code (e.g. that reads an int and matches it to a constructor).
  * non GADTs via regular generics
    * they will fail if an inner type isn't set up properly


raehik:
Dominic did a bunch of preliminary work and found this to be an issue over in branch [replaceConstantFolder](https://github.com/camfort/fortran-src/tree/replaceConstantFolder)

raehik:
I can't figure out how to derive `Data` for `SomeFKinded`. If I add a `Typeable` constraint on the existential type, changing

```haskell
data SomeFKinded k ft
  = forall (fk :: k). (SingKind k, SingI fk) => SomeFKinded (ft fk)
```

into

```haskell
data SomeFKinded k ft
  = forall (fk :: k). (SingKind k, SingI fk, Typeable fk) => SomeFKinded (ft fk)
```

then I get issues in my `... => Binary (SomeFKinded k ft)` instance, because now `get` has to magic a `Typeable` constraint out of nowhere. I can do that for `SingI` using `withSingI`, but I can't see any similar function for `Typeable`.

raehik:
There are some known issues with this changeset, namely poor support for handling inter-dependent constants-- but this was previously not handled, so I'm merging with the intent of improving the situation over time. In the meanwhile, we can swap out the expression evaluator in fortran-vars's syntax analysis with this (which was heavily inspired).

Issue #256, opened 2023-02-23
Unable to parse a procedure containing comments in its parameter list


uNouss:
Hello everyone, I have an example of code in fortran77 in which in the declaration of the parameters of procedures over several lines using line continuation, there are comments as on the following example.

```fortran
C234567890123456789012345678901234567890123456789012345678901234567890123
      subroutine say_hello(
C == fname: first name ==
     & fname,
C == lname: last name ==
     & lname)
         character fname*(*), lname*(*)
         write(*,*) 'Hello ' , fname , ', ' , lname
      end subroutine say_hello
      program main
         character *10 fname, lname
         fname = 'John'
         lname = 'Doe'
         call say_hello(fname, lname)
      end program main
C----------------------------------------------------------------------^
```
I can compile it, so parse it, with `gfortran` but not with `fortran-src` with the following error message.

You can see it here to reproduce the bug
[![asciicast](https://asciinema.org/a/hTRheOjBL6cpoYTtVFM8w4Eft.svg)](https://asciinema.org/a/hTRheOjBL6cpoYTtVFM8w4Eft)

RaoulHC:
Looks like it does parse fine if you use the legacy flag, i.e. `fortran-src -v77l test.f`.

I'm not entirely sure why it's confined to the legacy parser, but there's logic in the fixed form lexer in `alexGetByte` for skipping comments that are before a continuation line. It might be that this isn't actually allowed in the Fortran 77 standard but I will check. If not this should be made more general to the different fortran 77 parsers.

uNouss:
@RaoulHC Thank you for the answer.

yes indeed with the legacy flag in addition, the parsing works.

RaoulHC:
Looks like the 77 standard explicitly says this should be allowed in section 3.2 [here](https://wg5-fortran.org/ARCHIVE/Fortran77.html), so I'll look at making this behaviour a bit more general.

uNouss:
Okay thanks

RaoulHC:
FYI I was having a slightly deeper look at the Fortran standard and there's a few things in your example that are actually non-standard and so the Fortran 77 parser fails to compile: underscore in a symbolic name, symbolic names that are longer than 6 characters and `end subroutine id`. So a strict Fortran77 standard parser should actually still fail.

Given how little old Fortran is actually strictly standard compliant I wonder if we should change the default to Fortran 77 legacy for `.f` files. Any thoughts on this @raehik or @dorchard?

[...]

PR #257, opened 2023-02-23
Loosen legacy restriction for inline comments

RaoulHC:
These are explicitly in the Fortran 77 standard.

Issue raised in #256.

### DIFF
--- a/src/Language/Fortran/Parser/Fixed/Lexer.x
+++ b/src/Language/Fortran/Parser/Fixed/Lexer.x
@@ -944,7 +944,7 @@ alexGetByte ai
   -- Skip the continuation line altogether
   | isContinuation ai && _isWhiteInsensitive = skip Continuation ai
   -- Skip the newline before a comment
-  | aiFortranVersion ai == Fortran77Legacy && _isWhiteInsensitive
+  | aiFortranVersion ai >= Fortran77 && _isWhiteInsensitive
   && isNewlineCommentsFollowedByContinuation ai = skip NewlineComment ai
   -- If we are not parsing a Hollerith skip whitespace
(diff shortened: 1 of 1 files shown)


raehik:
Cheers, nice catch.

To confirm, there are two things we're doing here:

  * removing F77Legacy restriction for comment lines in-between continuation lines -- this is in the F77 standard
  * removing F77Legacy restriction for `!` inline comments -- don't believe this is in F77, but one of the many common extensions

fortran-src is generally pretty stubborn about syntax and standards, but I agree with you that we should be lenient here, especially with comments (no semantic value). LGTM!

---

Unrelated, the comment for

```haskell
  -- Skip the newline before a comment
  | _isWhiteInsensitive && isNewlineCommentsFollowedByContinuation ai
  = skip NewlineComment ai
```

threw me. It's actually detecting the former case, and skips that (comment) line.

raehik (approved):

RaoulHC:
Hmmm actually maybe we just want the first change in that case, indeed it looks like using `!` for comments isn't in the standard at all! Maybe that change should be removed and we can make the legacy parser the default for `.f` files as I suggested in #256? I think it could be useful to have a strict 77 standard parser, but the majority of users probably actually want the legacy parser.

raehik:
I'm fairly ambivalent on the `!` change. I'm definitely for your `.f` default parser change suggestion -- opened an issue at #259 .

Ah, I now note that this change loosens further than `Fortran77` -- the FORTRAN 66 parser also receives these changes. We should be able to use `Ord FortranVersion` to check for "FORTRAN 77 or newer" to catch all the F77s. (It'd be so much nicer if our lexers were composable...! aleeeeex!!)

PR #274, opened 2023-07-18
Postfix boz in free form lexer and generalised include inlining

RaoulHC:
Forgot to upstream these two patches that I wrote when helping some people with their Fortran 90 codebase.

First is a super small change for support for postfix boz notation which I encountered.

Second is for handling include inlining in the general case. It only worked on Fortran 77 before because that was all we cared about, but it probably should have been made to work for all parsers to begin with. I think it's sufficiently tested by the existing Fortran 77 tests, but let me know if you think it needs some more.

### DIFF
--- a/src/Language/Fortran/Parser/Free/Lexer.x
+++ b/src/Language/Fortran/Parser/Free/Lexer.x
@@ -51,9 +51,9 @@ $hash = [\#]
 @label = $digit{1,5}
 @name = $letter $alphanumeric*
 
-@binary = b\'$bit+\'
-@octal  = o\'$octalDigit+\'
-@hex    = z\'$hexDigit+\'
+@binary = b\'$bit+\' | \'$bit+\'b
+@octal  = o\'$octalDigit+\' | \'$octalDigit+\'o
+@hex    = [xz]\'$hexDigit+\' | \'$hexDigit+\'[xz]
 
 @digitString = $digit+
 @kindParam = (@digitString|@name)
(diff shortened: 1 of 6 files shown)


src/Language/Fortran/Parser/Free/Lexer.x
```diff
@@ -51,9 +51,9 @@ $hash = [\#]
 @label = $digit{1,5}
 @name = $letter $alphanumeric*
 
-@binary = b\'$bit+\'
-@octal  = o\'$octalDigit+\'
-@hex    = z\'$hexDigit+\'
+@binary = b\'$bit+\' | \'$bit+\'b
+@octal  = o\'$octalDigit+\' | \'$octalDigit+\'o
+@hex    = [xz]\'$hexDigit+\' | \'$hexDigit+\'[xz]
```
raehik, line 56: I was curious and confirmed: this is what the fixed lexer looks like. Synchronizing these is sensible, since I did work earlier on "early re-parsing" for these regexes (see `AST.Literal.Boz`). Which means if users want to, they can still catch the nonstandard syntax.

raehik (approved):
Superb. I like extending include inlining to the rest of the parsers. Existing tests are good enough for me. The BOZ lexing change is good too since we re-parse those regexes later (and nonconforming syntax can be caught).

I'll delay merging for a bit so I can run this by @dorchard . Thanks Raoul!

dorchard:
Yes LGTM.

PR #275, opened 2023-07-26
Pretty printer additions

RaoulHC:
Some fixes and some changes to accepted standards.


raehik:
Cheers! I see a number of `| v == FortranX = ...` guards (including the two you adapted) that use `tooOld`. Which is inaccurate for the check they perform. I added another error handler for those cases.

PR #276, opened 2023-07-28
Fix data group and pointer span

RaoulHC:
Previously it grabbed the span across the keyword and the reversed list which meant it only got the span until the end of first list, even if each list had the correct span.


raehik:
Ack, this seems like a really easy mistake to make. On a quick text search, I find another likely culprit:

https://github.com/camfort/fortran-src/blob/2e05dbecdae31d30e1498748aa400b5874bc53c1/src/Language/Fortran/Parser/Free/Fortran2003.y#L979-L981

And in the same file, an example of the correct form (matching your fix):

https://github.com/camfort/fortran-src/blob/2e05dbecdae31d30e1498748aa400b5874bc53c1/src/Language/Fortran/Parser/Free/Fortran2003.y#L616-L618

Would be nice if this sort of mistake was harder to make -- for now I'll ping an issue to track unchecked cases.

This change LGTM.

raehik:
I clarified what happens in #277 since the exact implementation is a bit hidden (chains through a few instances). Merging this.

Issue #282, opened 2024-06-14
Add Ord instances


Boarders:
It would be helpful to add derived Ord instances to almost all types so they can be used in standard containers such as `Data.Set`. I'd be happy to do this if the idea seems palatable.

raehik:
Thanks @Boarders . Sensible and adds a bit of convenience, I see no major reason not to. (Compilation time is a concern, but it doesn't seem to have an outrageous impact.) We should note that AST `Ord` instances don't guarantee any specific behaviour, other than being a valid instance. I've pushed a PR over at #283 .

Boarders:
Thanks, that is very helpful! I'm not sure if this would make more sense as a request after the modules have been split up slightly (if compilation times are the worry), but it may make sense at some point to also add `Hashable` and `NFData` (though I would say those are less important)

dorchard:
https://github.com/camfort/fortran-src/pull/283 is now merged for a new release 0.16.1. Does that satisfy the issue at hand suitably?

raehik:
Closing as resolved as I'm confident #283 is what @Boarders had requested.

Boarders:
That's correct - thanks both!

PR #283, opened 2024-06-15
Derive `Ord` instance for all core AST types

raehik:
Requested in #282 . Note that `Eq` was already derived.


raehik:
The AST module is very large with lots of automatic deriving already taking place. Deriving `Ord` as well will increase compilation time. Should be kept in mind. (It'd be nice to record compilation time per module, wonder if there's a tool for that.)

dorchard:
I think this is a good idea and worth the extra time taken, which is not vast. I timed a difference of about 5s (97s vs 102s) on builds between master and this branch.

dorchard:
I have rebased off `master`

PR #285, opened 2024-09-03
Some helpers for working with constant expression evaluator

dorchard:
Adds:
```
fromConstInt :: FValue -> Maybe Integer
fromConstReal :: FValue -> Maybe Double
```
which adds helper deconstructor on the `FValue` representation (which is useful for CamFort)


src/Language/Fortran/Repr/Value/Machine.hs
```diff
@@ -18,3 +20,14 @@ data FValue = MkFScalarValue FScalarValue
...
+fromConstInt :: FValue -> Maybe Integer
+fromConstInt (MkFScalarValue (FSVInt a)) = Just $ withFInt a
+fromConstInt _          = Nothing
+
+fromConstReal :: FValue -> Maybe Double
+fromConstReal (MkFScalarValue (FSVReal (FReal4 a))) = Just $ floatToDouble a
+  where
+    floatToDouble :: Float -> Double
+    floatToDouble = realToFrac
+fromConstReal (MkFScalarValue (FSVReal (FReal8 a))) = Just $ a
```
raehik, line 33: `fromConstReal` is currently partial. Probably want a fallback case like `fromConstReal _ = Nothing`

raehik (changes requested):
`fromConstReal` is partial due to a missed fallback case.

I'm happy to keep adding this sort of handy evaluator util a bit arbitrarily until we have enough to properly sort into a module of their own.

Re changelog: Let me know if I can help with the (Hackage) release.

src/Language/Fortran/Repr/Value/Machine.hs
dorchard: Thanks. Fixed in 5bd841f

raehik (approved):

PR #286, opened 2024-09-03
Localise module maps

dorchard:
A problem arose in CamFort with the location of declarations being misreported to the wrong file (https://github.com/camfort/camfort/issues/185). It turns out this comes down to the structure of mod file analysis, where all the `use`d modules get their name space inserted into the module map and so it cannot be distinguished whether a variable was declared in this module or an imported one. This PR fixes this by adding information into the `NameType` data structure about whether a variable is `Local` or `Imported`. A module map can then be 'localised' with `localisedModuleMap` which removes from a mod map any imported declarations. This fixes the bug in CamFort. 
An accompanying test was added in `ModFileSpec`

The `genUniqNameToFilenameMap :: ModFiles -> M.Map F.Name String` applies this localisation.


dorchard:
(note this is rebased off https://github.com/camfort/fortran-src/pull/285)

app/Main.hs
```diff
@@ -50,11 +50,16 @@ import qualified Language.Fortran.Parser.Free.Lexer  as Free
...
+showVersion :: String
+showVersion = "0.16.0"
+
 main :: IO ()
 main = do
   args <- getArgs
   (opts, parsedArgs) <- compileArgs args
   case (parsedArgs, action opts) of
+    (paths, ShowMyVersion) -> do
+      putStrLn $ "fortran-src version: " ++ showVersion
```
raehik, line 62: (We should have a release guide that instructs updating the version string shown here. I'll start a wiki page later.)

src/Language/Fortran/Util/ModFile.hs
```diff
@@ -217,6 +217,9 @@ decodeModFiles' = fmap (map snd) . decodeModFiles
 combinedModuleMap :: ModFiles -> FAR.ModuleMap
 combinedModuleMap = M.unions . map mfModuleMap
 
+localisedModuleMap :: FAR.ModuleMap -> FAR.ModuleMap
+localisedModuleMap = M.map (M.filter (not . FA.isImported . snd))
```
raehik, line 221: this threw me for a bit thanks to the deep type synonyms (`ModuleMap` storing a `ModEnv` which itself is actually a `Map`) -- could perhaps clarify what it's doing in a comment

raehik (approved):
Looks good to me. Some of the pre-existing type synonyms I find hard to follow and might warrant extra documenting, but it's a minor nit.

src/Language/Fortran/Util/ModFile.hs
raehik: that is, I saw `map . M.filter` and my first thought was "isn't this just a filter operation", but we're actually filtering the maps inside our top-level map

dorchard: Added a comment!

dorchard: Thanks!

PR #289, opened 2024-09-13
Store source names for declMaps in mod files

dorchard:
CamFort did not seem to have a way to get the source names for local declarations inside of program units. This made various analysis a bit annoying. The information is now provided here.

Along the way there is also some improvement to the evaluator for exponents in commit https://github.com/camfort/fortran-src/pull/289/commits/d14608003478010b1725207e2198f8513e86351e and https://github.com/camfort/fortran-src/pull/289/commits/f0c07b34fba0bd72bafe7b5d6848ef354d8d468a


dorchard:
This appears good to go now

PR #290, opened 2024-09-15
try to permit building with GHC 9.6

raehik:


raehik:
This adds a bunch of redundant imports for GHC <9.6. I propose this doesn't matter. I only know how to sidestep it with CPP (which is a very silly idea here). My usual solution is to add a comment on the import line so it shows up in the warning, which I've done here.

PR #292, opened 2024-09-25
Improve reporting, mostly around type information but also source spans

dorchard:
This PR does two things:
1. Improves the reporting of type information, e.g.
previously we'd get things like:

```
_area_of_circle_1                - Function
area_of_circle_area_2           Real 4 Variable
area_of_circle_pi_4             Real 4 Parameter
area_of_circle_r_3              Real 4 Variable
main_area_6             Real 4 Variable
main_r_5                Real 4 Variable
```
now we get:

```
4:12     r              Real 4 Variable
4:15     area           Real 4 Variable
11:4     area_of_circle          - Function
12:27    pi             Real 4 Parameter
13:28    r              Real 4 Variable
14:16    area           Real 4 Variable
```

2. This PR deals with the long standing issues that source span column numbers are always 1 bigger than they should be! This is now fixed in the `Show` instance (as fixing it deeper down causes issues with the way the lexer automata works).

### DIFF
--- a/src/Language/Fortran/Util/Position.hs
+++ b/src/Language/Fortran/Util/Position.hs
@@ -26,7 +26,7 @@ instance Binary Position
 instance NFData Position
 
 instance Show Position where
-  show (Position _ c l _ _) = show l ++ ':' : show c
+  show (Position _ c l _ _) = show l ++ ':' : show (c - 1)
 
 initPosition :: Position
 initPosition = Position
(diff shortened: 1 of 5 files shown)


src/Language/Fortran/Util/Position.hs
```diff
@@ -26,7 +26,7 @@ instance Binary Position
 instance NFData Position
 
 instance Show Position where
-  show (Position _ c l _ _) = show l ++ ':' : show c
+  show (Position _ c l _ _) = show l ++ ':' : show (c - 1)
```
raehik, line 29: Could you add your 2nd point in the PR and the PR number #292 here as a comment?

raehik (changes requested):

raehik (approved):

Issue #303, opened 2025-02-15
support for deprecated or proprietary constructs? `record //` and `structure //`.


smoothdeveloper:
I'm trying out the parser, but it doesn't handle some of the idioms in the code I'm dealing with (there will be a stream of this), here is a first example array declaration:

```fortran
record /myRecordType/ variableName(*)
```

relying on such a structure definition:

```fortran
structure /myRecordType/
  integer fieldA
  real fieldB
end structure
```

I'm interested in learning to extend the parser but wonder how such non-standard extensions would be integrated?

Also, my haskell skills are pretty low, I'm trying to understand how the .x & .y files are processed, and would like to improve the code to have a verbose command line flag, that would trace things as the lexer and parser are processing; which would help figuring out the flow of things.

smoothdeveloper:
@RaoulHC, thanks for all the pointers (erm, targets)!

I was just looking in the older lexer and got hopeful seeing both `record` and `structure`, giving `-v 77l` helps but I'm still getting `lexing failed`, this time it outputs a `AlexInput {...}` stuff I can actually diagnose a bit better.

I'll read up on `alex` and `happy` to gain better understanding.

Now, the codebase I'd like to process is using a mix of idioms used indiscriminately, those constructs are used in free form files.

Is there a sense that we don't want the constructs supported in later lexers, due to added complexity?

Would you recommend me to create my own custom lexer / parser pair, by copying the 2003 and integrating the bits I encounter and are in the 77l grammar?

Do you suggest a particular approach if we'd like the tool to have `--verbose` flag that outputs information about the processing as it is ongoing? I think it would be useful to the user and the maintainers as well.

RaoulHC:
Ah yes it doesn't support that, perhaps because it wasn't ever encountered in free form files. The lexing of fixed and free form is very different, so you'd have to add support for those tokens to the free form lexer, and then perhaps you can just add a combination [here](https://github.com/camfort/fortran-src/blob/main/src/Language/Fortran/Parser.hs#L151-L160) that parses the 77 syntax tree after a free form lexer pass. There were a lot of newer features added to the old parser (dealt with a lot of fortran 90 and newer constructs being used in fixed form legacy code), so even if you've got a horrific mix of standards it should mainly work.

smoothdeveloper:
@RaoulHC thanks.

I've started something in #305, note that in the codebase I'm looking at, I also have things like `[allocatable](:)`.

[...]

PR #305, opened 2025-02-15
WIP: Qualified fortran version

smoothdeveloper:
I'm starting work to allow having `CompilerOption` list to describe the introduced `QualifiedFortranVersion`, which will enable supporting `DecStructure` option.

Looking for maintainers feedback if this is right approach?

As of now, I'd like to get things to work back, which should be done by adjusting Parser.hs and maybe some more things in Main.hs, and later revise the new symbol names, which right now, may not be what we ultimately want.

If a maintainer want to look at the remaining type check error and suggest the right approach, as it is not clear to me at this stage.

Once this compile, I have a bit more work in the lexer to also have the `TRecord` token and maybe other things.

related: #303

### DIFF
--- a/README.md
+++ b/README.md
@@ -143,6 +143,16 @@ We support the latest recommended version of Cabal (as of 2021-09-17, Cabal 3.4)
 cabal build
 ```
 
+You can leverage cabal to initialize a ghci session, and call the main function like such:
+
+```shell
+cabal repl
+
+:load app/Main.hs
+
+:main --version
+```
+
 ### Testing
 Unit tests are stored in `test`. Run with `stack test` or `cabal test`.
 
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -16,7 +16,7 @@ import System.Directory
 import System.FilePath
 import Text.PrettyPrint.GenericPretty (pp, pretty, Out)
 import Text.Read (readMaybe)
-import Data.List (sortBy, intercalate, isSuffixOf)
+import Data.List (sortBy, intercalate, isSuffixOf, union)
 import Data.Ord (comparing)
 import Data.Char (toLower)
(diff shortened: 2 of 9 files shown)


smoothdeveloper:
Ok, I succeeded in flowing the entry point options down to the parsers, now I'm facing some challenge to understand what I should do in the free form lexer, to support the `/MyStructName/` that comes right after the `structure` keyword.

In the fixed form lexer I see:

https://github.com/camfort/fortran-src/blob/b8a490fca804f5f512195417af369834629b6710/src/Language/Fortran/Parser/Fixed/Lexer.x#L232

and the free form lexer:

https://github.com/camfort/fortran-src/blob/b8a490fca804f5f512195417af369834629b6710/src/Language/Fortran/Parser/Free/Lexer.x#L319

https://github.com/camfort/fortran-src/blob/b8a490fca804f5f512195417af369834629b6710/src/Language/Fortran/Parser/Free/Lexer.x#L694-L699

I'm unsure how to proceed, when I've parsed `structure` should I change context so as to process the "/" differently? 

I'd assume this flow:

`[TStructure, TSlash, TId "MyStructName", TSlash, ...]`

What do those contexts mean:

https://github.com/camfort/fortran-src/blob/b8a490fca804f5f512195417af369834629b6710/src/Language/Fortran/Parser/Monad.hs#L33-L40

I'll look more at this test and try to add one for free form:

https://github.com/camfort/fortran-src/blob/b8a490fca804f5f512195417af369834629b6710/test/Language/Fortran/Parser/Fixed/LexerSpec.hs#L270-L274

smoothdeveloper:
Wondering if we want to separate the QualifiedFortranVersion, which changes a bunch of signatures for the "already instanciated parsers" and the rest of the change I started working on (supporting option in QualifiedFortranVersion to parse the structure / record declarations in >= Fortran 90).

PR #310, opened 2025-04-29
[WIP] Code generator random Fortran programs

dorchard:
This is a WIP PR to provide a technique for automatically enumerating valid Fortran programs using QuickCheck style generation. The aim is to provide a facility for easily generating sample programs which can be used for testing other tools, e.g., refactoring tools and compilers. This leverages the machinery of QuickCheck's `Arbitrary` class.

Typing and scoping are the main challenges. Looking into work such as [Making Random Judgments: Automatically Generating
Well-Typed Terms from the Definition of a Type-System (ESOP 2015)](http://users.eecs.northwestern.edu/~baf111/random-judgments/random-judgments-esop15.pdf)
Also relevant: https://dl.acm.org/doi/pdf/10.1145/3363562
See also CSmith: https://users.cs.utah.edu/~regehr/papers/pldi11-preprint.pdf

- [x] Simple value generator
- [x] Declarations
- [ ] Source spans and positions (that do not need to be accureate)
- [x] Well-typed assignments
- [x] Well-typed values
- [ ] Well-typed expressions
- [ ] More statements
- [ ] Program units

Usage:
```
fortran-src --generate N -o DIR
```
generates `N` programs in the `DIR` directory.

I have also put a script in the top-level directory for comparing `gfortran` and `ifort`:
```
python compare_compilers.py out
```

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -46,6 +46,7 @@ import Language.Fortran.Analysis.Renaming
 import qualified Language.Fortran.Parser as Parser
 import qualified Language.Fortran.Parser.Fixed.Lexer as Fixed
 import qualified Language.Fortran.Parser.Free.Lexer  as Free
+import Language.Fortran.Generate (generatePrograms)
 
 programName :: String
 programName = "fortran-src"
@@ -212,6 +213,12 @@ main = do
                            | otherwise    = B.replicate (maxLen - B.length line + 1) ' ' <> "!" <> nodeStr
                 B.putStrLn $ line <> suffix
         _ -> fail $ usageInfo programName options
+    (_, Generate n) -> do
+      let dir = fromMaybe "." (outputFile opts)
+      exists <- doesDirectoryExist dir
+      unless exists $ ioError $ userError $
+        "Output directory does not exist: " ++ dir
+      generatePrograms n dir
     _ -> fail $ usageInfo programName options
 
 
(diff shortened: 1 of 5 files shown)


ksromanov:
@dorchard 

Dominic, there is also a problem of an ecosystem for such a tool:

Suppose we write such a tool and target `fortran-src` only. Even at the very early stage of development the tool might reveal some problems, that are going to get fixed. However, to uncover more errors, the tool has to become more and more sophisticated. Since the errors will be fixed almost immediately (it is much easier to fix them than to find), the tool will become less useful with time, though getting more and more complex and hard to maintain. At some point we might end up with a tool, that is a state-of-art fuzzer that still can find nothing in `fortran-src`.

Therefore, we either should target other parsers/compilers as well, or CI to check for `fortran-src` regressions.

As an example, our simple fuzzy experiment was able to find a bug https://github.com/camfort/fortran-src/pull/251, however it was the only finding. So, at the moment it is completely useless.

dorchard:
This is a good point @ksromanov - I was thinking about generating programs to test other compilers and tools, not just fortran-src itself. That would at least extend its lifetime / value, i.e., even if it turns up no issues with tool X, it might do in another new tool Y. But looking also at more 'fuzzing' based approaches might be good to, i.e., parsing in a program, then generating a lot of programs based from it. Do you think that goes beyond the problem you foresee, at least somewhat?

nikolaushuber:
Might be of interest:
https://dl.acm.org/doi/full/10.1145/3647994

dorchard:
We could use the expression generator to test the two expression evaluator: https://github.com/camfort/fortran-vars/blob/main/src/Language/Fortran/Vars/Eval

dorchard:
Okay I've done function call synthesis using the classic inverted sequent calculus style function application rule-  just need to actually populate the function environment with intrinsics, and also to do something similar for binary and unary operators (since these are represented differently to function calls in the AST, but logically it'll be the same kind of synthesis approach as here).

PR #314, opened 2025-08-06
 Stack at LTS-23.27, Cabal builds under GHC 9.0 thru 9.12

mrd:
Modifications to stack.yaml and package.yaml to enable the above.


.github/workflows/ci.yml
```diff
@@ -21,7 +21,7 @@ jobs:
 
     # relative paths are relative to the project directory
     - name: Cache Stack build artifacts (user + project)
-      uses: actions/cache@v2
+      uses: actions/cache@v3
```
foxtran, line 24: Just wonder why not actions/cache@v4...

raehik (approved):
Brilliant. Thanks for keeping the Mac & Win builds up to date as well.
I was going to say it's a miracle our `singletons-{th,base}` usage stayed compatible, but inspecting the code that uses it (that I wrote), I don't think we use anything beyond the base definitions in `singletons`. We can probably prune those (somewhat bulky) dependencies.

PR #321, opened 2026-01-09
Store only the base name of a source file in the fsmod file

dorchard:
Previously `.fsmod` files were storing the file filename, which makes them not very portable when using a tool at a different relative directory to the fsmod (this was the underlying issue behind https://github.com/camfort/camfort/issues/190 in CamFort.

This simple fix stores only the files base name in the fsmod file which makes them more portable/referable from different directory contexts.


mrd:
I'm making a new camfort branch `fortran-src-0.17` since the API has changed for modfiles. You can see how it affects things there. https://github.com/camfort/camfort/pull/227

Issue #323, opened 2026-01-16
`fortran-src 0.16.9` looses continue statements


foxtran:
Considering the following Fortran 77 code:
```fortran
      PROGRAM TEST
      IMPLICIT NONE
      PARAMETER (MAXLAY=20)
      DIMENSION A(MAXLAY), B(MAXLAY), C(MAXLAY)
      INTEGER NLAY
      OPEN(UNIT=10)
      READ(10, *) NLAY
      DO 100 I = 1, NLAY
         READ(10, *) A(I), B(I), C(I)
  100 CONTINUE
      CLOSE(10)
      END
```

Reparsing of this looses end of cycle:
```
% fortran-src --fortranVersion=77 example.f -r
      program test
      implicit none
      parameter (maxlay = 20)
      dimension a(maxlay), b(maxlay), c(maxlay)
      integer nlay
      open (unit=10)
      read (10, *) nlay
      do 100 i = 1, nlay
        read (10, *) a(i), b(i), c(i)
      close (10)
      end
```

dorchard:
Yes added on that PR (you can run these specifically with `stack test --ta="--match \"labeled continue\""`)

dorchard:
This has been fixed and merged now. Are you happy for me to close the issue and we can cut another release? @foxtran

PR #324, opened 2026-01-22
Pretty-printer for Fortran 77 needs to re-interpret do blocks

dorchard:
This PR addresses issue #323 

Fortran 77 has unstructued do, e.g., with code like:
```
      DO 100 I = 1, NLAY
         READ(10, *) A(I), B(I), C(I)
     100 CONTINUE
```
fortran-src applies a grouping transformation internally which turns this into a do-block, throwing away the continue statement. However, when pretty-printing, we do not re-insert the continue statement, so get:
```
      DO 100 I = 1, NLAY
         READ(10, *) A(I), B(I), C(I)
```
We must re-insert such a continue statement otherwise the code will be invalid; note, there is no `end do` in Fortran 77 so the continue must go back in. Relatedly in >= Fortran 90  mode we seem to be failing to output an `end do` in this case too.

This PR adds this behaviour to the pretty printer, and also does something related in Fortran 90 mode when there is a labelled do with and end label. Tests added to check.

### DIFF
--- a/app/Main.hs
+++ b/app/Main.hs
@@ -138,7 +138,7 @@ main = do
       case actionOpt of
         Lex | version `elem` [ Fortran66, Fortran77, Fortran77Extended, Fortran77Legacy ] ->
           print $ Parser.collectTokens Fixed.lexer' $ initParseStateFixed "<unknown>" version contents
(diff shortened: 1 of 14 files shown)


dorchard (commented):
Great- looks good to go- thank you!

PR #327, opened 2026-05-17
Make `-Woperator-whitespace`-sane

sol:
I'm working on a code indexing system that currently trips over `$u`.

`-Woperator-whitespace` enforces more than what my system currently needs, but it is easy to enforce.

Before the code used used mix conventions, especially in regards to `"foo"<>show x<>"bar"` vs `"foo" <> show x <> "bar"`; now it is consistent everywhere.

Do you think this is an improvement in general?

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+/*.cabal linguist-generated=true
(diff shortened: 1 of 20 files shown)


dorchard:
Sure - this seems like a reasonable lint to me and I can't see any issue in the tests. Happy to merge this.

